PR #11, opened 2019-03-16
Support RelatedField specific kwargs in its __init__.

kalekseev:
RelatedField has several extra kwargs 

        many: bool = ...,
        allow_empty: bool = ...,
        queryset: Optional[QuerySet] = ...,
        html_cutoff: Optional[int] = ...,
        html_cutoff_text: Optional[int] = ...,

Not sure is it even possible to extend original Field.__init__ typings so I've copy pasted original Field typings.

### DIFF
--- a/rest_framework-stubs/relations.pyi
+++ b/rest_framework-stubs/relations.pyi
@@ -1,8 +1,8 @@
-from typing import Any, Dict, Iterable, List, Optional, Protocol, Sequence
+from typing import Any, Callable, Dict, Iterable, List, Mapping, Optional, Protocol, Sequence, Union
 
 from django.db.models import Model
 from django.db.models.query import QuerySet
-from rest_framework.fields import Field
+from rest_framework.fields import Field, _Validator
 from rest_framework.request import Request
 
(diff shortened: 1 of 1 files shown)


rest_framework-stubs/relations.pyi
```diff
@@ -32,6 +32,26 @@ class Option(Protocol):
...
+        default: Any = ...,
+        initial: Any = ...,
+        source: Union[Callable, str] = ...,
+        label: Optional[str] = ...,
+        help_text: str = ...,
+        allow_null: bool = ...,
+        validators: Optional[Sequence[_Validator]] = ...,
+        error_messages: Optional[Mapping[str, str]] = ...,
+        style: Optional[Mapping[str, Any]] = ...,
+        many: bool = ...,
```
mkurnikov: The only thing: 
please, add those additional parameters to the beginning of the parameter list, to be consistent with how other fields are defined. 

Otherwise looks great, thanks for the PR!

kalekseev: moved params to beginning of the list

PR #17, opened 2019-04-02
fix schema() decorator

rafales:
As per docs it's fine to pass `None`:

https://www.django-rest-framework.org/api-guide/views/#view-schema-decorator


mkurnikov:
Sorry, completely missed github notification. Thanks!

PR #18, opened 2019-06-10
Add `actions` parameter to `ViewSetMixin.as_view`.

cs-cordero:
The ViewSetMixin supports an actions positional parameter.
https://www.django-rest-framework.org/api-guide/viewsets/

https://github.com/encode/django-rest-framework/blob/master/rest_framework/viewsets.py#L49


rest_framework-stubs/viewsets.pyi
```diff
@@ -11,7 +11,7 @@ class ViewSetMixin(object):
     basename: str
     request: Any
     @classmethod
-    def as_view(cls, **initkwargs: Any) -> Callable: ...
+    def as_view(cls, actions: Optional[Dict[str, Any]] = None, **initkwargs: Any) -> Callable: ...
```
sobolevn: Isn't it `Optional[Dict[str, str]]`?

Why `str, str`? 
Searching trough the docs gave me this impression: https://github.com/encode/django-rest-framework/blob/master/rest_framework/viewsets.py#L6

Can you please point to any other use-cases?

mkurnikov: Yes, it seems that it's a mapping between http method and method name on the class, it should be `Dict[str, str]` or maybe `Mapping[str, str]` to be a bit more general.

jmolinski:
```suggestion
    def as_view(cls, actions: Optional[Dict[str, str]] = ..., **initkwargs: Any) -> Callable: ...
```

Hey @cs-cordero, can we move forward with this pull request? It'd be really nice to have it merged! You can simply apply the suggested change.

jmolinski: @mkurnikov `Mapping` could be unnecessarily broad. IMO the type should be `Dict[str, str]` unless someone explicitly requests to change it.

cs-cordero: My apologies on the delay.  Life got in the way, but I just pushed a new commit to move forward with this request.

PR #19, opened 2019-06-24
Add CurrentUserDefault

noamkush:


rest_framework-stubs/fields.pyi
```diff
@@ -39,6 +39,10 @@ class CreateOnlyDefault(object):
     def set_context(self, serializer_field: Field) -> None: ...
     def __call__(self) -> Any: ...
 
+class CurrentUserDefault(object):
+    def set_context(self, serializer_field: Field) -> None: ...
+    def __call__(self) -> Any: ...
```
sobolevn: @mkurnikov are we okay with `Any` in these cases?

mkurnikov: Sorry, I somehow missed it. Just add some `# TODO:` for later, with explanation why it's difficult to add return type now, and it's good to go.

noamkush: I pushed a change to make it `Union[AnonymousUser, User]` which is what [get_user](https://github.com/typeddjango/django-stubs/blob/335c5765c3960e3ebd296b65ebac27b09b1e44fe/django-stubs/contrib/auth/middleware.pyi#L8) returns. This assumes both `AuthenticationMiddleware` is used and there's no custom user model in use.

sobolevn:
This should had be merged now, sorry for the long delay!

P.S. facing the same issue right now:

```
server/rest_app/logics/serializers.py:21: error: Name 'fields.CurrentUserDefault' is not defined
```

sobolevn:
Docs: https://github.com/encode/django-rest-framework/blob/3c1428ff799e4fe19e1eb28ed63a7d54c6180f1d/rest_framework/fields.py#L283

mkurnikov:
@noamkush Sorry=) 

@sobolevn Actually, it should've been `Union[AnonymousUser, Model]`, no? Though it would create false-positives too, so I guess there's no difference...

sobolevn:
It is actually change is the most recent version of drf. 

Now it is:

```python
# TODO: fix when drf-stubs > 1.1 will be released
class _CurrentAppUserDefault(fields.CurrentUserDefault):  # type: ignore
    def __call__(self, serializer_field):
        return serializer_field.context['request'].user.link.profile
```

But again, we need to understand how to work with different versions of annotations.
Should we raise this question in mypy's bugtracker?

mkurnikov:
@sobolevn I've created one here https://github.com/python/typing/issues/693

PR #26, opened 2019-07-31
Add missing actions argument to ViewSetMixin.as_view signature

jmolinski:


jmolinski:
I just realized this is a duplicate of https://github.com/typeddjango/djangorestframework-stubs/pull/18.
That being said it uses `Dict[str, str]` instead of `Dict[str, Any]` as suggested in the comments in #18 so it's ready to be merged.

sobolevn:
Thanks @jmolinski!

Issue #12, opened 2019-03-18
Support serializer field with name "fields"


kalekseev:
Consider the case:

```python
 class FieldsSerializer(serializers.Serializer):
     ...
 
 
 class OtherSerializer(serializers.Serializer):
    fields = FieldsSerializer()
```

Mypy will complain with: `error: Incompatible types in assignment (expression has type "FieldsSerializer", base class "Serializer" defined the type as "Dict[str, Field]")`

MarcinWieczorek:
You're trying to shadow Serializer's property `fields`, I think you really shouldn't do that, internals might be relying on that

MarcinWieczorek:
imho it's a wontfix and should be closed.

sobolevn:
I agree. `fields` is defined in `Serializer`. The best practice is not to override names like that.

PR #32, opened 2019-08-27
Add more attribute to fields, fixes #31

### LINKED ISSUE #31: Typings for Field attributes
Currently a lot of Field's attributes are not exposed in types, eg:

```
error: "ChoiceField" has no attribute "field_name"
error: "Field" has no attribute "allow_blank"
error: "Field" has no attribute "allow_null"
error: "Field" has no attribute "default"; maybe "get_default"?
error: "Field" has no attribute "read_only"
error: "Field" has no attribute "required"
```

that cause a lot of headache when serializers modified dynamically.

kalekseev:
Fixes #31, this will give type errors when user defined field name clash with one of the attributes eg.:

```
    label = CharField()
```

but it seems like mypy don't provide a way to distinguish class attrs and instance vars https://github.com/python/mypy/issues/1097

Similar issue https://github.com/typeddjango/djangorestframework-stubs/issues/12 marked as won't fix.


rest_framework-stubs/fields.pyi
```diff
@@ -67,10 +79,10 @@ class Field:
         source: Union[Callable, str] = ...,
         label: Optional[str] = ...,
         help_text: str = ...,
-        allow_null: bool = ...,
```
kalekseev: reordered kwargs to repeat drf order

kalekseev:
tests on master fail with the same error on my machine

PR #38, opened 2019-10-09
Allow dependency on mypy 0.730

### LINKED ISSUE #35: Support for Mypy 0.730
Mypy 0.730 have been released and `setup.py` specifies `mypy>=0.720,<0.730` while django-stubs specifies `mypy>=0.730` in `1.2.0`. See: https://github.com/typeddjango/django-stubs/blob/master/setup.py#L24

DanLipsitt:
This is the latest mypy as of this time. Breaking changes only seem to involve the output format:
http://mypy-lang.blogspot.com/2019/09/mypy-730-released.html

Fixes #35.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ with open('README.md', 'r') as f:
     readme = f.read()
 
 dependencies = [
-    'mypy>=0.720,<0.730',
+    'mypy>=0.720,<=0.730',
     'django-stubs>=1.0.0',
     'typing-extensions'
 ]


henribru:
Note that 0.740 is out now, so would be nice to be compatible with that as well.

intgr:
@sobolevn This can be closed now, right? Since commit 141f64b9696a16711c6630b9f92044d46c291438.

sobolevn:
Yeap 👍 

Thanks!

PR #39, opened 2019-10-16
Accept sequence in paginate_queryset and return optional, fixes #33

### LINKED ISSUE #33: GenericAPIView.paginate_queryset incorrect return type
Specified as `List[Any]`, should be `Optional[List[Any]]`

https://github.com/encode/django-rest-framework/blob/master/rest_framework/generics.py#L165

```python
def paginate_queryset(self, queryset):
    """
    Return a single page of results, or `None` if pagination is disabled.
    """
    if self.paginator is None:
        return None
    return self.paginator.paginate_queryset(queryset, self.request, view=self)
```

kalekseev:
object_list can be a sequence here https://github.com/django/django/blob/5becb36a79a0c31c7cc9e76237239ebd0fdda18d/django/core/paginator.py#L31, none returned here https://github.com/encode/django-rest-framework/blob/3aa1089a6af6774c1e0b12977b85062f837998ca/rest_framework/generics.py#L169


kalekseev:
fixes https://github.com/typeddjango/djangorestframework-stubs/issues/33

kalekseev:
@sobolevn please review, it's simple and has references to the code

PR #41, opened 2019-10-19
Compat with mypy>=0.730 and django-stubs>=1.2.0

### LINKED ISSUE #40: Querysets are typed wrong since django-stubs 1.1.0
Since version 1.1.0 of django-stubs, `QuerySet` no longer takes two generic parameters. This is the reason the build is currently failing with a bunch of `"QuerySet" expects 1 type argument, but 2 given` errors. This change was introduced in https://github.com/typeddjango/django-stubs/commit/656105bab2d1da30b5273ab0e6272b74d4ed6236#diff-b2dbe3c4313d37ea08b37f9a3dea0e77, which also includes some other changes to query sets. https://github.com/typeddjango/django-stubs/commit/fc9a335dfde1d27e47c22a55e3c2e8921fd132b3 made some additional changes. Because of the other changes I'm not sure if the correct solution is to just drop the second generic parameter or if there's more to it.

sobolevn:
Closes #40 

Changes:
- Updated versions of dependencies
- Removed `DynamicListRoute` and `DynamicDetailRoute`, because they do not exist: https://github.com/encode/django-rest-framework/blob/master/rest_framework/routers.py
- Changed how `QuerySet` is typed
- Some minor tweaking of `.travis.yml` to make it consistent


tselepakis:
I think this PR also closes #35 .

PR #45, opened 2019-12-06
Correct the return type of BasePagination.get_paginated_response

brianhelba:
* The DRF docs state that the return type should be a Response:
  https://www.django-rest-framework.org/api-guide/pagination/#custom-pagination-styles
* Existing DRF subclasses also return a Response:
  https://github.com/encode/django-rest-framework/blob/90eaf51839e2bfc3ca897d1f87028ca3303aa097/rest_framework/pagination.py#L220


brianhelba:
@mkurnikov @sobolevn This is ready for review.

PR #50, opened 2019-12-13
Allow Mapping method names as field names

lundberg:
This PR addresses an issue where field names conflicts with method names of `Mapping`, which `Serializer` extends.

If a serializer has a field named `keys`, `values`, `items` or `get`, an **assignment** error is thrown by mypy.

**Error example**:
```py
# test.py
from rest_framework.serializers import Serializer, CharField


    items = CharField()  # [assignment] error
```
```console
$ mypy test.py
test.py:6: error: Incompatible types in assignment (expression has type "CharField", base class "Mapping" defined the type as "Callable[[Mapping[str, BoundField]], AbstractSet[str]]")
```

This is solved by dropping `Mapping` inheritance. DRF does not fully implement Mapping anyways.

### DIFF
--- a/rest_framework-stubs/serializers.pyi
+++ b/rest_framework-stubs/serializers.pyi
@@ -111,7 +111,7 @@ class SerializerMetaclass(type):
 
 def as_serializer_error(exc: Exception) -> Dict[str, List[ErrorDetail]]: ...
 
-class Serializer(BaseSerializer, Mapping[str, BoundField]):
+class Serializer(BaseSerializer):
     @property
     def fields(self) -> Dict[str, Field]: ...
     @property


lundberg:
Additionally, same goes for `Serializer` properties `data`, `errors` and `fields`.

If a serializer has a field named as one of those properties, an assignment error will also be thrown.

These are *harder* to solve, since it would affect typing when accessing the actual properties... thoughts welcome.

sobolevn:
Thanks a lot! Sorry for the delay.

sobolevn:
@lundberg can you please create a new issue about typing `data`, `error`, `fields`, etc?

lundberg:
> @lundberg can you please create a new issue about typing `data`, `error`, `fields`, etc?

Added issue #78

sobolevn:
Thanks a lot! 👍

PR #53, opened 2019-12-18
allow lists fo strings for ordering attribute

### LINKED ISSUE #52: CursorPaginator.ordering should allow a list of strings
Having something like this in the code...

``` python
class Pagination(pagination.CursorPagination):
    ordering = ("-created_at", "pk")
```

... throws a typing error:

```
 .../pagination.py:123: error: Incompatible types in assignment (expression has type "Tuple[str, str]", base class "CursorPagination" defined the type as "str")
        ordering = ("-created_at", "pk")
                    ^
```

  .
  

The [docs](https://www.django-rest-framework.org/api-guide/pagination/#configuration_2) say:

> * ordering = This should be a string, or list of strings, indicating the field against which... 

The [code](https://github.com/encode/django-rest-framework/blob/master/rest_framework/pagination.py#L820-L828) says:

``` python
class CursorPagination(BasePagination):
    ...
    def get_ordering(self, request, queryset, view):
        ...
        assert isinstance(ordering, (str, list, tuple)), (
            'Invalid ordering. Expected string or tuple, but got {type}'.format(
                type=type(ordering).__name__
            )
        )

        if isinstance(ordering, str):
            return (ordering,)
        return tuple(ordering)
```

frnhr:
closes #52


rest_framework-stubs/pagination.pyi
```diff
@@ -87,7 +87,7 @@ class CursorPagination(BasePagination):
     cursor_query_description: str = ...
     page_size: int = ...
     invalid_cursor_message: str = ...
-    ordering: str = ...
+    ordering: Union[str, List[str, ...], Tuple[str, ...]] = ...
```
sobolevn: It should be `Union[str, List[str], Tuple[str, ...]]`
Thanks for the fix!

frnhr: Yes... Fingers being quicker than brain... 🚀

mkurnikov: Can't it just be `Sequence[str]`?

frnhr: Yeah, it would work I think, except that there is this `assert` in the source that explicitly checks for these 3 types.

sobolevn: Yes, I agree with @frnhr here. There's an explicit runtime typecheck there.

PR #55, opened 2019-12-19
Added action attribute to ViewSetMixin

### LINKED ISSUE #54: ModelViewSet is missing action attribute
Already briefly mentioned in issue https://github.com/typeddjango/djangorestframework-stubs/issues/9

The `action` attribute is defined in [`ViewSetMixin.initialize_request`](https://github.com/encode/django-rest-framework/blob/master/rest_framework/viewsets.py#L131-L143), which is called from [`APIView.dispatch`](https://github.com/encode/django-rest-framework/blob/d985c7cbb999b2bc18a109249c583e91f4c27aec/rest_framework/views.py#L481-L488), AFAICS. The point is that, while it is possible to get a view that has no `action` attribute (by overriding `dispatch` as part of some nasty agenda :D ) in normal use any view that inherits form `ViewSetMixin` will have a `self.action`.

I think we should add a type annotation for this attribute in the stubs.

frnhr:
closes #54


fgblomqvist:
What's blocking this from getting merged?

sobolevn:
Thanks!

PR #58, opened 2020-01-07
Fix field type arguments

asfaltboy:
I came across these couple of issues when using ImageField in my project


kalekseev:
@sobolevn I think this merge should be reverted, because it conflicts with https://github.com/typeddjango/djangorestframework-stubs/pull/62

sobolevn:
@kalekseev does it? I have manually resolved conflicts, it seemed fine. Can you please highlight conflicting places?

kalekseev:
@sobolevn the master build is broken because now we have duplicate attributes in fields, check the master code, these are duplicates:
- https://github.com/typeddjango/djangorestframework-stubs/blob/master/rest_framework-stubs/fields.pyi#L147
- https://github.com/typeddjango/djangorestframework-stubs/blob/master/rest_framework-stubs/fields.pyi#L174

and many more.

I think all fields were already covered by https://github.com/typeddjango/djangorestframework-stubs/pull/62 so this merge should be reverted.

PR #59, opened 2020-01-22
Add/improve ViewSet instance attributes

intgr:
Documented at https://www.django-rest-framework.org/api-guide/viewsets/#introspecting-viewset-actions

### DIFF
--- a/rest_framework-stubs/viewsets.pyi
+++ b/rest_framework-stubs/viewsets.pyi
@@ -1,21 +1,28 @@
 from collections import OrderedDict
 from typing import Any, Callable, Dict, List, Optional
 
+from django.http.response import HttpResponseBase
+
 from rest_framework.request import Request
 
 from rest_framework import generics, mixins, views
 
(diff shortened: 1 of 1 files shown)


intgr:
Oops, pushed a fix for this. Turns out I can't change the `as_view()` method because otherwise it would conflict with one from django-stubs. Also removed `request` attribute whcih is already defined in `View` superclass.

intgr:
I'm really looking forward to this. :)

I use ViewSets and reference `self.action` in many places.

sobolevn:
I am not sure about when the release will happen though

sobolevn:
Thanks again! ;)

intgr:
Sorry, brown paper bag moment. I somehow omitted the `action` attribute. I think I was looking up whether it's nullable or not, and got sidetracked. Now I found that there are even more instance attributes.

sobolevn:
No problem, send one more PR.

PR #60, opened 2020-01-22
Make decorator functions transparent to Mypy

intgr:
By declaring return type as `-> Callable[[_VIEW], _VIEW]`, Mypy can infer
that the decorated view has also the same arguments and return type as
the original.

View functions are constrained to return HttpResponseBase (or any
subclass of it).

## Related issues
Analogous PR in django-stubs: https://github.com/typeddjango/django-stubs/pull/306


rest_framework-stubs/decorators.pyi
```diff
@@ -7,23 +9,22 @@ from rest_framework.renderers import BaseRenderer
 from rest_framework.schemas.inspectors import ViewInspector
 from rest_framework.throttling import BaseThrottle
 
-_ViewCallable = Callable
-_ViewDecorator = Callable[[Callable], _ViewCallable]
+_VIEW = TypeVar("_VIEW", bound=Callable[..., HttpResponseBase])
```
intgr: Note that this doesn't work if I make `_ViewDecorator` an alias for `Callable[[_VIEW], _VIEW]`. It only works if the `Callable[]` is directly the function's return type.

I used the variable name `_VIEW` instead of `_ViewCallable`, because TypeVar variables are usually in allcaps.

intgr:
Pushed an update with improved tests, cleaned up 1 unused import.

sobolevn:
Awesome feature!

PR #62, opened 2020-01-29
More precise *Field arguments and instance attributes

intgr:
Field constructors are quite inconsistent in Django REST Framework: some take position arguments, some don't. Some pass positional arguments to their parent class, others pass only keyword arguments. It took a while but I sorted through all the constructors and nailed down the precise behavior.

The stubs did not previously denote keyword-only arguments, and the order of positional arguments was slightly wrong in many places.

Also, added many instance attributes of *Field classes that were missing.

### DIFF
--- a/rest_framework-stubs/fields.pyi
+++ b/rest_framework-stubs/fields.pyi
@@ -1,11 +1,12 @@
 import datetime
 import decimal
 import uuid
+from decimal import Decimal
+from json import JSONEncoder
 from typing import (
     Any,
     Callable,
     Dict,
-    Generic,
     List,
     NoReturn,
     Optional,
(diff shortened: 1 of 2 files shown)


intgr:
Pushed a small update, I found that `ChoiceField.choices` is actually a property that's always an `OrderedDict`.

rest_framework-stubs/fields.pyi
```diff
@@ -1,11 +1,13 @@
 import datetime
 import decimal
 import uuid
+from collections import OrderedDict
```
sobolevn: Should we explicitly specify that it is `OrderedDict` and not just `dict`?
I guess that they have the same API.

I would be against this change.

intgr: Using `Dict[]` is fine. Pushed an update for this.

I suppose `OrderedDict` is on its way to deprecation anyway, now that CPython's regular dict is now order-preserving.

sobolevn:
Thanks!

intgr:
Thanks too, working on TypedDjango projects has been pleasant. :)

PR #64, opened 2020-01-31
Even more ViewSet instance attributes

intgr:
These attributes are not documented and were reverse engineered from DRF source code.

### DIFF
--- a/rest_framework-stubs/viewsets.pyi
+++ b/rest_framework-stubs/viewsets.pyi
@@ -1,5 +1,5 @@
 from collections import OrderedDict
-from typing import Any, Callable, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional, Tuple
 
 from django.http.response import HttpResponseBase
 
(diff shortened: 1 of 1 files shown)


rest_framework-stubs/viewsets.pyi
```diff
@@ -12,11 +12,18 @@ def _is_extra_action(attr: Any) -> bool: ...
...
 class ViewSetMixin(object):
+    # Classvars assigned in as_view()
     name: Optional[str]
     description: Optional[str]
     suffix: Optional[str]
     detail: bool
     basename: str
+    # Instance attributes assigned in view wrapper
+    action_map: Dict[str, str]
+    args: List[Any]
```
sobolevn: Its't it `Tuple[Any, ...]`?

intgr: Oh yes, you are right. Finally found the time to fix this.

intgr:
Rebased my branch. I see that #54 already added the `action` attribute :)

sobolevn:
Thanks again!

PR #68, opened 2020-04-09
Revert "Fix field type arguments"

sobolevn:
Reverts typeddjango/djangorestframework-stubs#58


sobolevn:
@kalekseev reverted.

sobolevn:
It was more complex than a simple revert.
`django-stubs@1.5` introduced new typing to `test.pyi` and it broke the inheritance for `drf`.
Now, these errors are ignored, because we cannot do anything else about it. That's how it is by design.

Issue #69, opened 2020-04-09
ModelSerializer missing Meta attribute


stevanmilic:
`ModelSerializer` should a have `Meta` attribute with `model`, since it cannot be initialized without it.

Is there any reason why it isn't already included?

I can make a PR with the fix.

sobolevn:
PRs are welcome! Thanks!

PR #70, opened 2020-04-09
Add `Meta` attribute to `ModelSerializer` (#69)

stevanmilic:
`ModelSerializer` cannot be instantiated withuot the `Meta` class set
and `model` defined on it.

### DIFF
--- a/rest_framework-stubs/serializers.pyi
+++ b/rest_framework-stubs/serializers.pyi
@@ -12,6 +12,7 @@ from typing import (
     NoReturn,
     MutableMapping,
     Iterator,
+    Union,
 )
 
 from rest_framework.fields import (
@@ -149,6 +150,13 @@ class ModelSerializer(Serializer):
     serializer_url_field: Type[RelatedField] = ...
     serializer_choice_field: Type[Field] = ...
     url_field_name: Optional[str] = ...
+    class Meta:
+        model: Type[Model]
+        fields: Optional[Union[Sequence[str], str]]
+        read_only_fields: Optional[Sequence[str]]
+        exclude: Optional[Sequence[str]]
+        depth: Optional[int]
+        extra_kwargs: Dict[str, Dict[str, Any]]
     def get_field_names(self, declared_fields: Mapping[str, Field], info: FieldInfo) -> List[str]: ...
     def get_default_field_names(self, declared_fields: Mapping[str, Field], model_info: FieldInfo) -> List[str]: ...
     def build_field(
--- a/test-data/typecheck/test_serializers.yml
+++ b/test-data/typecheck/test_serializers.yml
@@ -9,3 +9,29 @@
                 pass
         class SerializerC(SerializerA, SerializerB):
             pass
+
+- case: model_serializer_meta_attributes
+  main: |
+    from rest_framework import serializers
+
+    reveal_type(serializers.ModelSerializer.Meta.model) # N: Revealed type is 'Type[django.db.models.base.Model]'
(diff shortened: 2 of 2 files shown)


rest_framework-stubs/serializers.pyi
```diff
@@ -142,7 +143,16 @@ class ListSerializer(BaseSerializer):
 
 def raise_errors_on_nested_writes(method_name: str, serializer: BaseSerializer, validated_data: Any) -> None: ...
 
+class ModelSerializerMeta:
```
sobolevn, line 146: Can you please write a test for it?

stevanmilic: yea, sure

stevanmilic: what's the proper way to test this? should I just test the revealed types for the `ModelSerializer.Meta.*`?

stevanmilic: @sobolevn I've added the test for the above^. If you think I should do something additional, just point it out :)

stevanmilic: btw I was thinking, maybe it's better to define `Meta` as a class inside `ModelSerializer`, like this:
```python
class ModelSerializer(Serializer):
    class Meta:
        model: Type[Model]
        fields: Optional[Union[Sequence[str], str]]
        read_only_fields: Optional[Sequence[str]]
        exclude: Optional[Sequence[str]]
        depth: Optional[int]
        extra_kwargs: Dict[str, Dict[str, Any]]
    ...
```

test-data/typecheck/test_serializers.yml
```diff
@@ -8,4 +8,15 @@
             class Meta:
                 pass
         class SerializerC(SerializerA, SerializerB):
-            pass
\ No newline at end of file
+            pass
+
+- case: model_serializer_meta_attributes
```
sobolevn, line 13: Thanks a lot! This is a nice addition.

But, I was talking about some real serialiser definition that uses `ModelSerialiser` and `Meta.model` class.

Because I fear that `Meta` can use nominal subtyping and cause `mypy` to raise here.

stevanmilic: Okay, I've added another test. I'm not sure if it fits into test case you imagined.

I don't understand in which case `Meta` can use the nominal subtyping, can you elaborate on it?

stevanmilic: > Okay, I've added another test. I'm not sure if it fits into test case you imagined.
Oh yea, I get it because I used the `ModelSerializerMeta`, I've changed it to:
```python
class ModelSerializer(Serializer):
    class Meta:
        model: Type[Model]
        fields: Optional[Union[Sequence[str], str]]
        read_only_fields: Optional[Sequence[str]]
        exclude: Optional[Sequence[str]]
        depth: Optional[int]
        extra_kwargs: Dict[str, Dict[str, Any]]
    ...
```

test-data/typecheck/test_serializers.yml
```diff
@@ -8,4 +8,30 @@
...
+    from typing import Type
+
+    class TestSerializer(serializers.ModelSerializer):
+        class Meta:
+          model = User
+
+    def is_meta_model(serializer: Type[serializers.ModelSerializer]):
+        return bool(serializer.Meta.model)
+
+    is_meta_model(TestSerializer)
```
sobolevn, line 37: Let's `reveal_type` here that it is `bool` and we are done!

stevanmilic: should be good now

PR #83, opened 2020-07-21
Add generic type parameter to Serializer.validate fixes #82

### LINKED ISSUE #82: Make Serializer.validate generic?
Serializer.validate is currently untyped:

https://github.com/typeddjango/djangorestframework-stubs/blob/master/rest_framework-stubs/serializers.pyi#L124-L125

Is it possible to do something like

```py
T = TypeVar("T"):
class Serializer(BaseSerializer):
    def validate(self, attrs: T) -> T:
        return attrs
```

I'm experimenting with using [TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict) to declare explicit interfaces for the data that a serializer processes, and I can annotate an overridden method like so:

```py
class MySerializer(serializers.Serializer):
    name = serializers.CharField()

    class MySerializerData(TypedDict):
        name: str

    def validate(self, data: MySerializerData) -> MySerializerData:
        reveal_type(data["name"])  # str
        validated_data = super().validate(data) # Call to untyped function "validate" in typed contextmypy(error)
```

I'm just learning how Python typing works so please forgive my lack of knowledge!

Thanks!

jaredlockhart:
I'm not sure how to run the test suite to validate this?

### DIFF
--- a/rest_framework-stubs/serializers.pyi
+++ b/rest_framework-stubs/serializers.pyi
@@ -8,6 +8,7 @@ from typing import (
     Callable,
     Mapping,
     Type,
+    TypeVar,
     Tuple,
     NoReturn,
     MutableMapping,
@@ -121,7 +122,8 @@ class Serializer(BaseSerializer):
     def _readable_fields(self) -> List[Field]: ...
     def get_fields(self) -> Dict[str, Field]: ...
     def _read_only_defaults(self) -> Dict[str, Any]: ...
-    def validate(self, attrs):
+    T = TypeVar("T")
+    def validate(self, attrs: T) -> T:
         return attrs
     @property
     def data(self) -> ReturnDict: ...


jaredlockhart:
Oh nevermind I got pytest working.  Should I modify/add any tests?

rest_framework-stubs/serializers.pyi
```diff
@@ -121,7 +122,8 @@ class Serializer(BaseSerializer):
     def _readable_fields(self) -> List[Field]: ...
     def get_fields(self) -> Dict[str, Field]: ...
     def _read_only_defaults(self) -> Dict[str, Any]: ...
-    def validate(self, attrs):
+    T = TypeVar("T")
+    def validate(self, attrs: T) -> T:
```
sobolevn, line 126: Yes, let's test this. 

Here's an example: https://github.com/typeddjango/djangorestframework-stubs/blob/master/test-data/typecheck/test_serializers.yml

Goldziher:

PR #87, opened 2020-08-11
Update MyPy max pinned version to 0.782

sidmitra:
- The latest version of MyPy is 0.782 [here](https://pypi.org/project/mypy/#history)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ def find_stub_files(name):
 with open("README.md", "r") as f:
     readme = f.read()
 
-dependencies = ["mypy>=0.750,<0.780", "django-stubs>=1.3.0", "typing-extensions"]
+dependencies = ["mypy>=0.750,<=0.782", "django-stubs>=1.3.0", "typing-extensions"]
 
 setup(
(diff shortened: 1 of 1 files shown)


sidmitra:
I've been using the latest master with 0.782 for the past few weeks and everything seems to work correctly.

sidmitra:
Will try to take a look at the CI error later today. Not sure if it's related to the PR.

```
$ mypy -p rest_framework-stubs
rest_framework-stubs/request.pyi: error: Source file found twice under different module names: 'rest_framework-stubs.request' and 'rest_framework.request'  [misc]
Found 1 error in 1 file (checked 48 source files)
The command "mypy -p rest_framework-stubs" exited with 2.
```

sobolevn:
Yes, this is known bug in the latest `mypy`. That's why I had to pin it to the previous version. See https://github.com/python/mypy/issues/8944

sidmitra:
@sobolevn Thanks for the link, wasn't aware of that. I didn't come across any issues on our production codebase. I will take a deeper look, looks more involved though.

Goldziher:

PR #89, opened 2020-09-16
DynamicRoute namedtuple should not have a `mapping` attr

anentropic:
See definition here https://github.com/encode/django-rest-framework/blob/3.9.4/rest_framework/routers.py#L39

I am getting a `Missing positional argument "mapping" in call to "DynamicRoute"` error from mypy when trying to instantiate a DynamicRoute using these stubs currently

### DIFF
--- a/rest_framework-stubs/routers.pyi
+++ b/rest_framework-stubs/routers.pyi
@@ -19,7 +19,6 @@ class Route(NamedTuple):
 
 class DynamicRoute(NamedTuple):
     url: str
-    mapping: Dict[str, str]
     name: str
     detail: bool
     initkwargs: Dict[str, Any]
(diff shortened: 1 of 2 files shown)


anentropic:
the failure in Travis CI is:
```
1.46s$ black --check --line-length=120 setup.py scripts/ rest_framework-stubs/
would reformat /home/travis/build/typeddjango/djangorestframework-stubs/rest_framework-stubs/test.pyi
Oh no! 💥 💔 💥
1 file would be reformatted, 48 files would be left unchanged.
The command "black --check --line-length=120 setup.py scripts/ rest_framework-stubs/" exited with 1.
```

this doesn't look related to my change 🤷‍♂️

sobolevn:
Can you please format it anyway?

anentropic:
done 👍

sobolevn (approved):

sobolevn:
Thanks a lot!

PR #90, opened 2020-10-02
get_url implementation might return None

ticosax:
in case the object is not yet persisted in the database

https://github.com/encode/django-rest-framework/blob/master/rest_framework/relations.py#L332


sobolevn (approved):

sobolevn:
Thanks a lot!

PR #94, opened 2020-10-09
Sync, update and debug against DRF

Goldziher:
This PR is a major overhaul of the typings in this package. What it does:

1. It syncs and adds almost all missing data from the latest version of the DRF
2. It fixes a lot of the previous typings and updates these to be in sync with the DRF
3. It adds significant improvements to typings - especially using generics for fields, relations and serialisers
4. It adds testing against the DRF test suit
5. It fully debugs the typings of this package and all changes against this test suit (this took me more than 30 hours to do). 

What remains to be done:
1. finish fix travis


Goldziher:
@sobolevn I might need help updating and fixing the test suit. 

Also - any chance of having a direct chat?

rest_framework-stubs/permissions.pyi
```diff
@@ -1,16 +1,17 @@
...
 SAFE_METHODS: Sequence[str] = ("GET", "HEAD", "OPTIONS")
 
 class _SupportsHasPermission(Protocol):
-    def has_permission(self, request: Request, view: View) -> bool: ...
-    def has_object_permission(self, request: Request, view: View, obj: Any) -> bool: ...
+    def has_permission(self, request: Request, view: APIView) -> bool: ...
+    def has_object_permission(self, request: Request, view: APIView, obj: Model) -> bool: ...
 
 _PermissionClass = Union[Type[BasePermission], OperandHolder, SingleOperandHolder]
+_OperationClass = Union[AND, OR, NOT]
```
kalekseev, line 14: I think it a rare case but someone could define its own operators

rest_framework-stubs/permissions.pyi
```diff
@@ -1,16 +1,17 @@
...
 from rest_framework.request import Request
+from rest_framework.views import APIView
 
 SAFE_METHODS: Sequence[str] = ("GET", "HEAD", "OPTIONS")
 
 class _SupportsHasPermission(Protocol):
-    def has_permission(self, request: Request, view: View) -> bool: ...
-    def has_object_permission(self, request: Request, view: View, obj: Any) -> bool: ...
+    def has_permission(self, request: Request, view: APIView) -> bool: ...
+    def has_object_permission(self, request: Request, view: APIView, obj: Model) -> bool: ...
```
kalekseev, line 11: view may operate on any object not models only, so obj should be Any I think

rest_framework-stubs/permissions.pyi
Goldziher: Its possible, but than he or she will need to override the typing in only one place - I dont think we should use Any in such circumstances.

rest_framework-stubs/permissions.pyi
Goldziher: Thats true regarding views in general (non generic ones at least), but object permissions are model specific in DRF afaik.

kalekseev: No, only DjangoObjectPermissions class and its children are model specific.

rest_framework-stubs/permissions.pyi
kalekseev: I'm not sure is it possible in mypy but something like _OperationClass = Type[_SupportsHasPermision] would be optimal

rest_framework-stubs/permissions.pyi
Goldziher: ok, fixed

rest_framework-stubs/permissions.pyi
Goldziher: its possible, i can define this as a protocol probably.

Goldziher: silly me, it was already a protocol. Done.

Goldziher:
ok, done :)

sobolevn:
@Goldziher this great! Awesome work! I will review this today.

Goldziher:
Thanks @sobolevn 😃. 

Is it possible to become an official member of the org?

rest_framework-stubs/decorators.pyi
```diff
@@ -37,5 +21,65 @@ class MethodMapper(dict):
...
+    throttle_classes: Sequence[Union[BaseThrottle, Type[BaseThrottle]]]
+) -> Callable[[Callable], Callable]: ...
+def permission_classes(
+    permission_classes: Sequence[Union[BasePermission, Type[BasePermission]]]
+) -> Callable[[Callable], Callable]: ...
+def schema(view_inspector: Optional[Union[ViewInspector, Type[ViewInspector]]]) -> Callable[[Callable], Callable]: ...
+def action(
+    methods: Optional[
+        List[
+            Union[
```
sobolevn, line 61: Let's make a type-alias for this

sobolevn (changes requested):
Thanks!

sobolevn:
@Goldziher invited! Thanks a lot for your hard work!

rest_framework-stubs/decorators.pyi
Goldziher: will do :)

Goldziher:
@sobolevn  cheers!

rest_framework-stubs/decorators.pyi
Goldziher: done

rest_framework-stubs/decorators.pyi
```diff
@@ -21,19 +21,40 @@ class MethodMapper(dict):
...
+        Literal["post"],
+        Literal["delete"],
+        Literal["put"],
+        Literal["patch"],
+        Literal["trace"],
+        Literal["options"],
+    ]
+]
+
+MIXED_CASE_HTTP_VERBS = List[
```
sobolevn, line 36: Just to clarify this:
1. we use `MIXED_CASE_HTTP_VERBS` in case there's an existing public contant in the original package
2. we use `_MIXED_CASE_HTTP_VERBS` in case we need to declare our own stuff

Goldziher: Kk, fixed

mypy_drf_plugin/main.py
```diff
@@ -13,7 +13,7 @@
 def transform_serializer_class(ctx: ClassDefContext) -> None:
     sym = ctx.api.lookup_fully_qualified_or_none(fullnames.BASE_SERIALIZER_FULLNAME)
     if sym is not None and isinstance(sym.node, TypeInfo):
-        helpers.get_drf_metadata(sym.node)['serializer_bases'][ctx.cls.fullname] = 1
+        helpers.get_drf_metadata(sym.node)["serializer_bases"][ctx.cls.fullname] = 1
```
last-partizan, line 16: Let's stick to consistent style of quotes, this project is using single quotes.

sobolevn (approved):

sobolevn:
Thanks a lot!

rest_framework-stubs/generics.pyi
```diff
@@ -9,28 +9,30 @@ from rest_framework.request import Request
...
     lookup_url_kwarg: Optional[str] = ...
     filter_backends: Sequence[Type[BaseFilterBackend]] = ...
     pagination_class: Optional[Type[BasePagination]] = ...
-    def get_queryset(self) -> QuerySet: ...
-    def get_object(self) -> Any: ...
-    def get_serializer(self, *args: Any, **kwargs: Any) -> BaseSerializer: ...
-    def get_serializer_class(self) -> Type[BaseSerializer]: ...
+    def get_object(self) -> _MT_co: ...
+    def get_serializer(self, *args: Any, **kwargs: Any) -> BaseSerializer[Any, Any, Any, _MT_co]: ...
+    def get_serializer_class(self) -> BaseSerializer[Any, Any, Any, _MT_co]: ...
```
kalekseev, line 30: @Goldziher this must return `Type[BaseSerializer]` not `BaseSerializer`

rest_framework-stubs/generics.pyi
```diff
@@ -9,28 +9,30 @@ from rest_framework.request import Request
...
     def get_serializer_context(self) -> Dict[str, Any]: ...
-    def filter_queryset(self, queryset: QuerySet) -> QuerySet: ...
+    def filter_queryset(self, queryset: QuerySet[_MT_co]) -> QuerySet[_MT_co]: ...
     @property
     def paginator(self) -> Optional[BasePagination]: ...
-    def paginate_queryset(self, queryset: Union[QuerySet, Sequence]) -> Optional[List[Any]]: ...
+    def paginate_queryset(self, queryset: QuerySet[_MT_co]) -> Optional[Sequence[Any]]: ...
```
kalekseev, line 35: @Goldziher I would say the old typing were better, even if it named paginate *queryset* it actually can work with Sequence too

rest_framework-stubs/response.pyi
```diff
@@ -1,19 +1,23 @@
-from typing import Any, Dict, Mapping, Optional
+from typing import Any, Dict, List, Mapping, Optional, Union
 
 from django.template.response import SimpleTemplateResponse
+from rest_framework.request import Request
 
 class Response(SimpleTemplateResponse):
-    data: Optional[Any] = ...
+    data: Optional[Union[List[Dict[str, Any]], Dict[str, Any]]] = ...
```
kalekseev, line 7: should be Any, these data just passed to renderer user can define renderer that accepts arbitrary data, also your typing doesn't handle all the cases of valid json, for example we have case where `data: List[List[str]]`

rest_framework-stubs/decorators.pyi
```diff
@@ -1,29 +1,13 @@
...
-
-_VIEW = TypeVar("_VIEW", bound=Callable[..., HttpResponseBase])
-
-def api_view(http_method_names: Optional[Sequence[str]] = ...) -> Callable[[_VIEW], _VIEW]: ...
-def renderer_classes(renderer_classes: Sequence[Type[BaseRenderer]]) -> Callable[[_VIEW], _VIEW]: ...
-def parser_classes(parser_classes: Sequence[Type[BaseParser]]) -> Callable[[_VIEW], _VIEW]: ...
-def authentication_classes(authentication_classes: Sequence[Type[BaseAuthentication]]) -> Callable[[_VIEW], _VIEW]: ...
-def throttle_classes(throttle_classes: Sequence[Type[BaseThrottle]]) -> Callable[[_VIEW], _VIEW]: ...
-def permission_classes(permission_classes: Sequence[Type[BasePermission]]) -> Callable[[_VIEW], _VIEW]: ...
-def schema(view_inspector: Optional[Type[ViewInspector]]) -> Callable[[_VIEW], _VIEW]: ...
```
intgr, line 19: You undid the work that I did in #60, this MR seems reckless.

The result is that argument and return types of views are erased if they use any decorator here.

test-data/typecheck/test_decorators.yml
```diff
@@ -3,22 +3,17 @@
...
 -   case: api_view_bare_is_error
     main: |
         from rest_framework.decorators import api_view
         @api_view  # E: Argument 1 to "api_view" has incompatible type "Callable[[Any], Any]"; expected "Optional[Sequence[str]]"
         def view_func2(request): ...
--   case: api_view_incorrect_return
-    main: |
-        from rest_framework.decorators import api_view
-        @api_view()  # E: Value of type variable "_VIEW" of function cannot be "Callable[[Any], List[Any]]"
-        def view_func2(request) -> list: ...
```
intgr, line 24: I think this was a useful feature and test case: the ability to detect `@api_view` decorator on wrongly typed functions. Simply removed with no discussion.

intgr (commented):
There are lots of good improvements here, and the effort is very much appreciated.

But there are also several places where this PR made things worse than they used to be and it comes off as reckless. It's generally accepted that regressions are worse than bugs that were always there.

I noticed the type stubs no longer behaving the way they did earlier, then I saw this massive change in the code, and it took me an afternoon to only take one scan over all the changes, to find several issues with it. I'm sure nobody likes surprises like that?

In the future, please:
* Be mindful when you're changing code in a way that loses precision or removing code that may perform some useful function. If you're not sure why something was the way it was, look at the git history, or state your doubts in PR comments. Ping the original authors of that change, etc.
* Submit changes as multiple incremental pull requests instead of everything at once. If you have too many changes in a single PR, it likely won't be reviewed in depth at all.

intgr:
I've resolved most of these in #119 and will open a separate pull request for validator callbacks.

rest_framework-stubs/response.pyi
intgr: Agreed. After restoring decorators logic in #119, many `response` variable types became visible to mypy in the test suite, and this hint causes false positives.

I've reverted it to `Any` in #119.

PR #95, opened 2020-10-09
Updated package setup

Goldziher:
This PR makes substantial improvements to the package setup - introducing pre-commit and making our lives easier. I suggest it is reviewed only after the huge PR from which it is branched is merged, to make this simpler.


LICENSE.txt
```diff
@@ -1,8 +1,20 @@
-Copyright (c) Maxim Kurnikov.
-All rights reserved.
+Copyright (c) Maxim Kurnikov
```
sobolevn, line 1: Let's keep the license file as-is, please.

release.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
sobolevn, line 1: `#!/usr/bin/env bash`

sobolevn (changes requested):
Thanks! Let's rebase this one properly and I will review it once again. 👍

Goldziher:
this PR was remade, closing this one.

PR #96, opened 2020-10-10
updated package setup

Goldziher:
This PR updates the package setup:

1. update scripts and setup.py
2. add pre-commit
3. update travis config
4. update contribution and readme .md files
5. add stubgen script

Please review @sobolevn


Goldziher:
I'm proceeding to update the django-stubs package along the lines of this one, so if there are any issues lemme know pls.

setup.py
```diff
@@ -34,11 +34,14 @@ def find_stub_files(name):
...
     package_data={"rest_framework-stubs": find_stub_files("rest_framework-stubs")},
+    python_requires=">=3.6",
     classifiers=[
-        "Development Status :: 3 - Alpha",
         "License :: OSI Approved :: MIT License",
+        "Operating System :: OS Independent",
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
```
sobolevn, line 44: We don't actually run tests on `3.9`, don't we? https://github.com/typeddjango/djangorestframework-stubs/blob/master/.travis.yml#L10

Btw, yesterday I have moved from travis to github actions: https://github.com/typeddjango/pytest-mypy-plugins/tree/master/.github/workflows

Maybe we should also?

setup.py
```diff
@@ -16,10 +16,10 @@ def find_stub_files(name):
     return result
 
 
-with open("README.md", "r") as f:
+with open("README.md") as f:
     readme = f.read()
 
-dependencies = ["mypy>=0.782", "django-stubs>=1.6.0", "typing-extensions", "requests>=2.0.0", "coreapi>=2.0.0"]
+dependencies = ["mypy>=0.782", "django-stubs>=1.6.0", "typing-extensions>=3.7.2", "requests>=2.0.0", "coreapi>=2.0.0"]
```
sobolevn, line 22: Let's bump `mypy` to `0.790`

setup.cfg
```diff
@@ -1,18 +1,16 @@
...
-multi_line_output = 5
-wrap_length = 120
-
 [flake8]
-exclude =
-    drf-sources,
-    rest_framework-stubs,
-    test-data
+exclude = .*/
+select = F401, Y
```
sobolevn, line 3: Why do we need this line?

setup.cfg
```diff
@@ -1,18 +1,16 @@
...
-include_trailing_comma = true
-multi_line_output = 5
-wrap_length = 120
-
 [flake8]
-exclude =
-    drf-sources,
-    rest_framework-stubs,
-    test-data
+exclude = .*/
```
sobolevn, line 2: Let's use explicit excludes. This is really hard to read / find.

sobolevn (changes requested):
Thanks a lot!

setup.py
Goldziher: TBH we only need to run tests on 3.6 IMO, if it passes.it will pass for later versions because the changes are only typings.

As for github actions - sure, but maybe make this in a separate PR after this one?

setup.py
Goldziher: Will do

setup.cfg
Goldziher: Ok

setup.cfg
Goldziher: We don't, I'll remove it

setup.py
Goldziher: so this is not currently possible because of this:
```
ERROR: django-stubs 1.6.0 has requirement mypy<0.790,>=0.782, but you'll have mypy 0.790 which is incompatible.
```
So I'll just keep it at >=0.782

sobolevn: ok

setup.py
sobolevn: Ok!

sobolevn: > TBH we only need to run tests on 3.6 IMO

It is better to run tests on all python versions we officially support. Better safe than sorry.

sobolevn (approved):

Issue #97, opened 2020-10-11
Make serializers less strict


kalekseev:
Hi @Goldziher, in your https://github.com/typeddjango/djangorestframework-stubs/pull/94/files you made a lot of improvements to the serializer's types, so I decided to try it in one of my projects and provide some feedback.

First of all thanks for your work it looks like you spent a lot of time working on that pr, the main problem that I see right now is that types become too specific, for example serializer generic signature https://github.com/typeddjango/djangorestframework-stubs/blob/master/rest_framework-stubs/serializers.pyi#L94  `class BaseSerializer(Generic[_VT, _DT, _RP, _IN], Field[_VT, _DT, _RP, _IN]):` with such signature I have to provide 4 types or not provide types at all so all generics become `Any`.
Here why I think `_VT`, `_DT` and `_RP` are not that useful
- `_VT` - used in initial and default attributes, you can easily provide that in code when needed
```
class S(Serializer):
     initial: InitialType
     default: DefaultType

     def __init__(..., initial: IntitialType, defautl: DefaultType,...):
```
- `_DT` - data that seriailzer will validate, it should be typed as `Any` or maybe `Union[dict, list]`, because main use case is that you accept anything as data and want to validate that it valid input

- `_RP` - representation method result, not sure why we should care about that type, it mostly internally used inside `def data` so it's better to have `data` return specific type instead of Any, anyway no reason to make it generic, you can easily override method return signature if you need it.

Basically when I work with serializer I care about these things:
- instance type
- result of the serialization `serializer.data`
- result of the validation `serializer.validated_data`

Of all of that cases I would argue that only instance type worth to make generic, because it used in signatures of a lot of methods, while you can always override result signature for `data` or `validated_data` methods when needed. So I would change signature to `Seriailizer(Generict[_IN])` until mypy has support for default values for generic eg. `Serailizer(Generic[_IN, _DT = Any, ...])`.

Another problem is that [these methods](https://github.com/typeddjango/djangorestframework-stubs/pull/94/files#diff-28c49e448e38ae9aaf7452c3747ff1dfR132-R134) + `def validate` works with OrderedDict, first of all that should be a `dict` because the ordering of the dict is not important for these methods, second is that user can write `serializer(many=True)` and these methods now must work with `List[dict]` instead of dict. So I would change their signature back to:
```
    def update(self, instance: Model, validated_data: Any) -> Any: ...
    def create(self, validated_data: Any) -> Any: ...
[...]

PR #99, opened 2020-10-11
made serializer typings simpler

Goldziher:
PR introducing the changes suggested by @kalekseev in #97


rest_framework-stubs/serializers.pyi
```diff
@@ -213,21 +204,22 @@ class ListSerializer(
...
+    def validate(self, attrs: Any) -> Any: ...
     @property
     def data(self) -> ReturnList: ...
     @property
     def errors(self) -> ReturnList: ...
 
 def raise_errors_on_nested_writes(method_name: str, serializer: BaseSerializer, validated_data: Any) -> NoReturn: ...
 
-class ModelSerializer(Serializer, BaseSerializer[_MT, Mapping[str, Any], Dict[str, Any], _MT]):
+class ModelSerializer(Serializer, BaseSerializer[_MT]):
```
sobolevn, line 215: Let's add a bunch of tests for this type! It would be super helpful in the future.

sobolevn:
@kalekseev I would love to hear your thoughts.

rest_framework-stubs/serializers.pyi
```diff
@@ -144,19 +141,16 @@ def as_serializer_error(exc: Exception) -> Dict[str, List[ErrorDetail]]: ...
 class Serializer(
     BaseSerializer[
         Union[_MT, Mapping[str, Any]],
```
kalekseev, line 143:
```suggestion
        Union[_IN, Mapping[str, Any]],
```
_MT is bound to Model so it should be used only with ModelSerializer

rest_framework-stubs/serializers.pyi
```diff
@@ -87,31 +87,28 @@ ALL_FIELDS: str = ...
 
 _MT = TypeVar("_MT", bound=Model)  # Model Type
 _IN = TypeVar("_IN")  # Instance Type
-_DT = TypeVar("_DT")  # Data Type
-_VT = TypeVar("_VT")  # Value Type
-_RP = TypeVar("_RP")  # Representation Type
 
-class BaseSerializer(Generic[_VT, _DT, _RP, _IN], Field[_VT, _DT, _RP, _IN]):
+class BaseSerializer(Generic[_IN], Field[Any, Any, Any, _IN]):
```
kalekseev, line 91: If we are going with all these generic params in Field I would move _IN to first place to make it uniform with Serializer

rest_framework-stubs/serializers.pyi
kalekseev: No need to make this generic this specific, let's use just _IN, someone can create subclass that will work with arbitrary objects not models or dict

rest_framework-stubs/serializers.pyi
```diff
@@ -171,9 +165,6 @@ class Serializer(
 
 class ListSerializer(
     BaseSerializer[
-        List[Any],
-        Union[MultiValueDict, List[Any]],
-        List[Dict[str, Any]],
         List[Dict[str, Any]],
```
kalekseev, line 168: same problem here, no need to make such constraints on user, for example I can pass QuerySet as instance, while you strictly require List of dicts for some reason.

kalekseev:
Serializers can be instantiated with instance = None, so it must be Optional, e.g.:
```
class Serializer(Generic[_IN]):
    def __init__(self, instance: Optional[_IN], ...):
```

rest_framework-stubs/serializers.pyi
kalekseev: also why in init `instance: Union[MultiValueDict, Sequence[Any]] = ...,`?

rest_framework-stubs/serializers.pyi
Goldziher: No time now...

rest_framework-stubs/serializers.pyi
Goldziher: I disagree.

1. Model serializer is a specialist class for models.
2. We do not need to account for what users might do, but for the drf
3. We need proper type checking

rest_framework-stubs/serializers.pyi
Goldziher: Its not used as often in field

rest_framework-stubs/serializers.pyi
Goldziher: This is based on the DRF tests and in codes comments.

Goldziher: changed

Goldziher:
Ok, all comments addressed in one way or another. I do not have time to work on this now for the remainder of the week. I would like to see this released though so I can actually use the stuff Ive worked on, so please consider proceeding @sobolevn

sobolevn (approved):

sobolevn:
Let's have a bit of time for this to vent, before releasing it to the public. Since we don't practice `-rc` or anything similar.
You can now use https://github.com/typeddjango/djangorestframework-stubs/commit/c5dbc028662748cc0119477f766db2dc190c2246 as a reference to install it from `git`.

Thanks a lot for your hard work! 👍

Issue #42, opened 2019-10-20
Support for singleton types in unions requires `empty` to be Enum


kalekseev:
[`__init__`](https://github.com/encode/django-rest-framework/blob/f98b6f35770c6cf7dd964d0fa7b052f6f86d5253/rest_framework/serializers.py#L109) and [`run_validation`](https://github.com/encode/django-rest-framework/blob/f98b6f35770c6cf7dd964d0fa7b052f6f86d5253/rest_framework/serializers.py#L419) methods have default value for data `=empty` so the data type is `Union[None, dict, Type[empty]]`:

```
def run_validation(self, data: Union[None, dict, Type[empty]]=empty):
    if data is None:
         ...  # data is None
    elif data is empty:
         ...  # data is Type[empty]
    else:
         ...  # data is Union[dict, Type[empty]]
```

Desired behavior:

```
def run_validation(self, data: Union[dict, None, Empty]=empty):
    if data is None:
         ...  # data is None
    elif data is empty:
         ...  # data is empty
    else:
         ...  # data is dict
```

Currently [DRF define empty](https://github.com/encode/django-rest-framework/blob/f98b6f35770c6cf7dd964d0fa7b052f6f86d5253/rest_framework/fields.py#L40) as:

```
class empty:
    pass
```

in order to discriminate union to dict we need it to be [enum](https://www.python.org/dev/peps/pep-0484/#support-for-singleton-types-in-unions):

```
class Empty(Enum):
    token = 0
empty: Final = Empty.token
```

**Solutions**

1) Define Empty as enum in stubs
```
from mypy import TYPE_CHECKING


if TYPE_CHECKING:
    from serializers.fields import Empty   # linters may complain that fields.py doesn't have Empty
else:
    class Empty:
        pass
```

2) Try to convince rest framework folks to change empty to enum
3) Try to convince rest framework folks to introduce some module like `types.py` where we can place our utility types like that one so user will be able to import it as `from rest_framework.types import Empty`. (would be useful for SerializerDataType concept I described here https://github.com/typeddjango/djangorestframework-stubs/issues/36#issuecomment-542766519)

PR #101, opened 2020-10-12
Made types less specific

kalekseev:
Also fixed some bugs, @sobolevn, @Goldziher please review.


rest_framework-stubs/fields.pyi
```diff
@@ -29,10 +31,15 @@ from django.core.files.base import File
...
-from rest_framework.serializers import BaseSerializer
 from typing_extensions import Literal
 
-class empty: ...
+from rest_framework.serializers import BaseSerializer
+
+class _Empty(Enum):
+    sentinel = 0
+
+empty: Final = _Empty.sentinel
```
kalekseev, line 41: Rationale https://github.com/typeddjango/djangorestframework-stubs/issues/42

sobolevn (changes requested):
Thanks a lot! 👍 

I really want to improve the testing part of this project. Because we now have lots of regressions due to the lack of type-tests. Can you please add one?

rest_framework-stubs/generics.pyi
```diff
@@ -27,12 +28,12 @@ class GenericAPIView(views.APIView, UsesQuerySet[_MT_co]):
...
     def get_object(self) -> _MT_co: ...
     def get_serializer(self, *args: Any, **kwargs: Any) -> BaseSerializer[_MT_co]: ...
-    def get_serializer_class(self) -> BaseSerializer[_MT_co]: ...
+    def get_serializer_class(self) -> Type[BaseSerializer[_MT_co]]: ...
     def get_serializer_context(self) -> Dict[str, Any]: ...
     def filter_queryset(self, queryset: QuerySet[_MT_co]) -> QuerySet[_MT_co]: ...
     @property
     def paginator(self) -> Optional[BasePagination]: ...
-    def paginate_queryset(self, queryset: QuerySet[_MT_co]) -> Optional[Sequence[Any]]: ...
+    def paginate_queryset(self, queryset: Union[QuerySet[_MT_co], Sequence]) -> Optional[Sequence[Any]]: ...
```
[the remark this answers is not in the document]
kalekseev: Why not?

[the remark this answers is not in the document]
kalekseev: The better option is to have two overrides one for queryset another for sequence, then we will have better typed result, but the goal of that pr was to fix typings not improve them so result type was preserved.

rest_framework-stubs/fields.pyi
```diff
@@ -79,6 +86,8 @@ _RP = TypeVar("_RP")  # Representation Type
 class SupportsToPython(Protocol):
     def to_python(self, value: Any) -> Any: ...
 
+_DefaultInitial = Union[_VT, Callable[[], _VT], None, _Empty]
+
 class Field(Generic[_VT, _DT, _RP, _IN]):
```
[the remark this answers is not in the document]
kalekseev: Will add when have another chunk of free time to work on this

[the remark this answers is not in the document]
sobolevn: Thank you! 🙏

rest_framework-stubs/generics.pyi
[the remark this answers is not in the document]
sobolevn: I was asking about styling:
- `queryset: Union[QuerySet[_MT_co], Sequence]`
vs 
- `queryset: Union[QuerySet[_MT_co], Sequence[Any]]`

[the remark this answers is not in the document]
kalekseev: ah, okay, I just reverted that line to the old version so missed that one

rest_framework-stubs/fields.pyi
```diff
@@ -79,6 +85,8 @@ _RP = TypeVar("_RP")  # Representation Type
 class SupportsToPython(Protocol):
     def to_python(self, value: Any) -> Any: ...
 
+_DefaultInitial = Union[_VT, Callable[[], _VT], None, _Empty]
```
Goldziher, line 88: Empty is not something the user is expected to pass externally, but is a fallback valiue user internally by the init method. As such this is wrong to put it inside the type alias here.

rest_framework-stubs/fields.pyi
```diff
@@ -538,8 +546,8 @@ class ChoiceField(Field[str, Union[str, int, Tuple[Union[str, int], Union[str, i
         read_only: bool = ...,
         write_only: bool = ...,
         required: bool = ...,
-        default: Union[str, int, Callable[[], str], Callable[[], int]] = ...,
-        initial: Union[str, int, Callable[[], str], Callable[[], int]] = ...,
+        default: Union[_Empty, str, int, Callable[[], str], Callable[[], int]] = ...,
```
Goldziher, line 549: not using the type alias here?

rest_framework-stubs/fields.pyi
```diff
@@ -572,8 +580,8 @@ class MultipleChoiceField(
         read_only: bool = ...,
         write_only: bool = ...,
         required: bool = ...,
-        default: Union[Sequence[str], Sequence[int], Callable[[], Sequence[str]], Callable[[], Sequence[int]]] = ...,
-        initial: Union[Sequence[str], Sequence[int], Callable[[], Sequence[str]], Callable[[], Sequence[int]]] = ...,
+        default: Union[_Empty, Sequence[str], Sequence[int], Callable[[], Sequence[str]], Callable[[], Sequence[int]]] = ...,
```
Goldziher, line 583: same

rest_framework-stubs/fields.pyi
```diff
@@ -598,8 +606,8 @@ class FilePathField(ChoiceField):
         required: bool = ...,
         read_only: bool = ...,
         write_only: bool = ...,
-        default: Union[str, int, Callable[[], str], Callable[[], int]] = ...,
-        initial: Union[str, int, Callable[[], str], Callable[[], int]] = ...,
+        default: Union[_Empty, str, int, Callable[[], str], Callable[[], int]] = ...,
```
Goldziher, line 609: same

rest_framework-stubs/fields.pyi
```diff
@@ -644,8 +652,8 @@ class ImageField(FileField):
         read_only: bool = ...,
         write_only: bool = ...,
         required: bool = ...,
-        default: Union[File, Callable[[], File]] = ...,
-        initial: Union[File, Callable[[], File]] = ...,
+        default: Union[_Empty, File, Callable[[], File]] = ...,
```
Goldziher, line 655: well same in all other places :)

rest_framework-stubs/generics.pyi
```diff
@@ -27,12 +28,12 @@ class GenericAPIView(views.APIView, UsesQuerySet[_MT_co]):
     pagination_class: Optional[Type[BasePagination]] = ...
     def get_object(self) -> _MT_co: ...
     def get_serializer(self, *args: Any, **kwargs: Any) -> BaseSerializer[_MT_co]: ...
-    def get_serializer_class(self) -> BaseSerializer[_MT_co]: ...
+    def get_serializer_class(self) -> Type[BaseSerializer[_MT_co]]: ...
```
Goldziher, line 31: does this method in fact return a class ? or does it expect a serializer instance?

rest_framework-stubs/reverse.pyi
```diff
@@ -1,21 +1,21 @@
 from typing import Any, Callable, Mapping, Optional, Sequence
 
-from rest_framework.request import Request
+from django.http import HttpRequest
```
Goldziher, line 3: why this change? It does not appear to be correct lookint at the code.

Goldziher (commented):
All in all nice stuff, I left some questions. I am not sure about the changes to reverse especially.

rest_framework-stubs/fields.pyi
Goldziher: interesting. did you raise this with the DRF folk?

kalekseev: No, because django core devs are very conservative about adding typings to their codebase also DRF right now is mostly in bugfixes state, so I don't have enough willpower to push that change.

rest_framework-stubs/fields.pyi
kalekseev: If user won't pass it as argument it doesn't mean that it shouldn't be included into type signature, type signature **must** include all valid values, example where that useful:

```
def __init__(self, ..., default: _DefaultInitial,...):
    if default is None or iscallable(default):
      ...
    else:
      ... <- here default will have type Union[_VT, _Empty], otherwise user would think that he checked everything and default is _VT

File: rest_framework-stubs/fields.pyi (hunk 5 above)
Comment (kalekseev) [reply rest_framework-stubs/fields.pyi:549]: not sure what's the point to write aliases for such cases, we can't use _DefaultInitial alias in this case

File: rest_framework-stubs/generics.pyi (hunk 9 above)
Comment (kalekseev) [reply rest_framework-stubs/generics.pyi:31]: class

File: rest_framework-stubs/reverse.pyi (hunk 10 above)
Comment (kalekseev) [reply rest_framework-stubs/reverse.pyi:3]: show how it can break the code please, Request is subclass of the HttpRequest

Comment (Goldziher) [reply rest_framework-stubs/reverse.pyi:3]: Thats not a valid reason. It can break the code because request is not httrequest and the DRF deals with request and not http request.

Comment (kalekseev) [reply rest_framework-stubs/reverse.pyi:3]: show how this code will break if I pass HttpRequest into these functions

File: rest_framework-stubs/fields.pyi (hunk 4 above)
Comment (Goldziher) [reply rest_framework-stubs/fields.pyi:88]: Maybe, but DRF doesn't expect user to pass it in. This is a non standard edge case.

Comment (kalekseev) [reply rest_framework-stubs/fields.pyi:88]: how it's edge case? if you call serializer without default argument in init you will have default=empty, even if it edge case I see zero reason why empty shouldn't be in the union

File: rest_framework-stubs/reverse.pyi (hunk 10 above)
Comment (sobolevn) [reply rest_framework-stubs/reverse.pyi:3]: For me it looks like `HttpRequest` can be used: https://github.com/encode/django-rest-framework/blob/master/rest_framework/reverse.py

Comment (Goldziher) [reply rest_framework-stubs/reverse.pyi:3]: @kalekseev - i asked you why you introduced this change, your answer "show how it break...". This is a poor answer. if there is no good reason to change the typing here, then dont please.

Comment (sobolevn) [reply rest_framework-stubs/reverse.pyi:3]: @Goldziher this is a pretty valid question. It  is about variance, since `Request` is a sub-type of `HttpRequest`.

Btw, we encourage people to contribute to our repos. Not discourage them to. 
So, any valid contributions: bugfixes, improvements, docs, etc are welcome 🙂

Comment (Goldziher) [reply rest_framework-stubs/reverse.pyi:3]: ofc, but is this an improvement? If there is a viable reason for doing this, e.g. typing this with Request causes an issue etc. than its fine, but if its just because of reasons then I am against this. 

In fact one of the most persistent issues I have with version 1.2 is that HttpRequest is used in multiple places where the DRF in fact uses Request, this causes MyPy to complain and I have to type ignore valid code.

The DRF uses Request, not HttpRequest, and Request has properties that are not available on HttpRequest. So again - is there a valid reason for making this change here?

Comment (Goldziher) [reply rest_framework-stubs/reverse.pyi:3]: one more thing @kalekseev  - i do encourage and appreciate your efforts (and knowledge)! These review comments are just code review - not person in any way.

Comment (kalekseev) [reply rest_framework-stubs/reverse.pyi:3]: @Goldziher relax man, I'm too old to care about someone words on the internet. Unfortunately code review is a hard work actually, I do them a lot and know that to make an effective review I need to understand the problem, load full context into my head and make research if I don't know some topic. So what I see from your comments and code is that you try to give a very specific type for all the things but actually in type theory this works another way, you want to give the least specific types for function arguments because the more types function accepts the more valuable it is. For example if function accepts list and just iterate on it, you could use argument type List but tomorrow you will find that you need same function for Set so it's better to make your function argument `Iterable`. In this particular case `reverse` function can work with `HttpResponse` without problem, it doesn't use any `Response` specific attributes or anything else so why would you want to constraint user to use `reverse` only with `Response` while function happily accepts HttpResponse and all its subtypes?

Comment (Goldziher) [reply rest_framework-stubs/reverse.pyi:3]: Fair enough, thanks for you explaining your PoV

**Review by Goldziher — APPROVED** [review]

**Comment by MarcinWieczorek** [comment]
I don't know how should the tests look for other parts, but I managed to test the `FloatField`.
Does this look alright? I'm not sure if the FloatField should allow integers, this patch says it does with code from this PR.
```patch
diff --git a/tests/typecheck/test_fields.yml b/tests/typecheck/test_fields.yml
index c1bc3fe..78d911a 100644
--- a/tests/typecheck/test_fields.yml
+++ b/tests/typecheck/test_fields.yml
@@ -65,3 +65,11 @@
 
         TimeField('', [], True, True, True, time(hour=1), time(hour=1), 'src', 'l', 'ht', {}, {}, [], True)
         TimeField('', [], True, True, True, time(hour=1), time(hour=1), 'src', 'l', 'ht', {}, {}, [], True, 1)  # E: Too many arguments for "TimeField"
+
+-   case: field_argument_types
+    main: |
+      from rest_framework.fields import FloatField
+      FloatField(min_value=1)
+      FloatField(max_value=1)
+      FloatField(min_value=1.0)
+      FloatField(max_value=1.0)
```

Goldziher:
Can this be merged?
@kalekseev can you resolve the conflicts?

MarcinWieczorek:
Now it merges. Are you going to include my test? sobolevn upvoted it.

rest_framework-stubs/fields.pyi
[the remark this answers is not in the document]
kalekseev: Added requested test

kalekseev:
> Now it merges. Are you going to include my test? sobolevn upvoted it.

Included, with small modification for compactness.

sobolevn (approved):
Thanks a lot!

palfrey:
Any chance of a new release with these changes in?

intgr:
I'd very much like to see a release that supports mypy 0.800 (or mypy 0.810, which seems like will be released soon).

What's missing from mypy 0.800 support again?

sobolevn:
@intgr mypy 0.800 introduced new explicit re-export rules. 

Now:
- This is an explicit re-export: `from a import b as b`
- This is not: `from a import b as c`

And we need to make sure that we meet this new rule.

rest_framework-stubs/fields.pyi
intgr: @kalekseev I had the same question and I don't understand the answer.

I think we **can** use
```python
default: _DefaultInitial[Union[str, int]] = ...,
initial: _DefaultInitial[Union[str, int]] = ...,
```

I opened PR #152, please respond there if you see any issues with my approach.

intgr:
I created a follow-up PR to convert more fields to the `_DefaultInitial[]` type alias: #152

rest_framework-stubs/fields.pyi
kalekseev: @intgr I don't remember my thoughts on that but most probably I just missed the ability to pass union as generic param, so your pr is on point

PR #114, opened 2020-12-13
Changed APIView.headers to Dict[str, str]

### LINKED ISSUE #76: APIView.headers type is missing
This line show that there's `headers` field in `APIView` class: https://github.com/encode/django-rest-framework/blob/5ce237e00471d885f05e6d979ec777552809b3b1/rest_framework/views.py#L494

Currently it is missing.

Related #74 and #75

MarcinWieczorek:
closes #76


sobolevn:
Context:
- https://github.com/encode/django-rest-framework/blob/bb133522efaf6ce3ae8fdf1dec6cd79566cfd166/rest_framework/views.py#L494
- https://github.com/encode/django-rest-framework/blob/bb133522efaf6ce3ae8fdf1dec6cd79566cfd166/rest_framework/views.py#L439

One question: is it `Dict[str, str]` or `Dict[str, Optional[str]]`? Is there any special `None` usage?

MarcinWieczorek:
That pop assigns None if `Vary` isn't in the dict and is used only for the `if` right below, therefore it doesn't count as `None` usage. I'm not aware of any other uses.

MarcinWieczorek:
I only found a similar case here: https://github.com/encode/django-rest-framework/blob/master/rest_framework/views.py#L456
None is used for flow control and is nullchecked before assignment. All other uses (as in `grep -re headers`) are either on strings constants or formatted strings as such: `headers['Retry-After'] = '%d' % exc.wait`

sobolevn (approved):
Thanks!

MarcinWieczorek:
Great, thanks for quick response.
I'll use this moment to remind you of my PR in django-stubs that has been idling since March :)

sobolevn:
@MarcinWieczorek oh, sorry! I might have missed it. I will check it out right now.

PR #117, opened 2020-12-15
Simplify Literal declarations

intgr:
A single `Literal[]` decalaration may contain multiple values, no need
to wrap it inside `Union[]`.


sobolevn (approved):
Good catch!

PR #119, opened 2021-01-12
Restore functionality lost in version 1.3.0

### LINKED ISSUE #112: Wrong types while using Model query in Response
I have the following code:

```python
class ListAssetsView(APIView):

    def get(self, request: Request) -> Response:
        return Response(Asset.objects.all().values(), status=status.HTTP_200_OK)

```

That fails with:
```
Argument 1 to "Response" has incompatible type 
"ValuesQuerySet[Asset, TypedDict({'id': Optional[UUID], 'name': str})]"; 
expected 
"Union[List[Dict[str, Any]], Dict[str, Any], str, None]"
```

To typing work, I need to convert the values list as: 
```python
result = [dict(val) for val in Asset.objects.filter(is_in_stock=True).values()]
return Response(result, status=status.HTTP_200_OK)
```

I there a way to make the types work without converting the ValuesQuerySet/TypedDict to List/Dict?

intgr:
Fixes for regressions introduced in #94 (commit cf9d492ec01f99e9a3fa89558aa1fbf1a749bc22)

* Functions wrapped with decorators (`@api_view` etc) don't lose their argument and return types.
* `Optional[]` added to some Field arguments that accept `None` value.
* Type hint for `BaseParser.parse()`.
* Structural types like `AsView` and `ViewSetAction`, of which there aren't any instances, declared as `Protocol`.
* Type of `Response.data` reverted to `Any` as otherwise it causes false positives (fixes #112; also https://github.com/typeddjango/djangorestframework-stubs/pull/94#discussion_r502972134)
* Field validators will be handled in #120.


intgr:
Pinging @Goldziher as the author of original PR #94 and @kalekseev who helped to review it.

rest_framework-stubs/parsers.pyi
```diff
@@ -8,6 +8,8 @@ from rest_framework.renderers import JSONRenderer
 _Data = TypeVar("_Data")
 _Files = TypeVar("_Files")
 
+_Validator = Callable[[Any], None]
```
kalekseev, line 11: unused?

kalekseev (approved):

rest_framework-stubs/parsers.pyi
intgr: Ah yeah, leftover. Will open separate pull request for this.

rest_framework-stubs/decorators.pyi
```diff
@@ -9,17 +10,19 @@ from rest_framework.throttling import BaseThrottle
 from rest_framework.views import APIView, AsView  # noqa: F401
 from typing_extensions import Literal
 
+_VIEW = TypeVar("_VIEW", bound=Callable[..., HttpResponseBase])
```
sobolevn, line 13: Do you prefer `_VIEW` over `_View`?

sobolevn (approved):
Thanks a lot!

rest_framework-stubs/decorators.pyi
intgr: Just out of habit, no strong preference. I see there are already typevars `_Data` and `_Files` so I can change it. to `_View`.

sobolevn (approved):
There's a missing test:

```
__________________________ api_view_incorrect_return ___________________________

/home/travis/build/typeddjango/djangorestframework-stubs/tests/typecheck/test_decorators.yml:25: 

E   pytest_mypy_plugins.utils.TypecheckAssertionError: Invalid output: 

E   Expected:

E     main:2: error: Value of type variable "_VIEW" of function cannot be "Callable[[Any], List[Any]]" (diff)

E   Actual:

E     main:2: error: Value of type variable "_View" of function cannot be "Callable[[Any], List[Any]]" (diff)

E   

E   Alignment of first line difference:

E     E: ... of type variable "_VIEW" of function cannot be "Callable[[Any], List...

E     A: ... of type variable "_View" of function cannot be "Callable[[Any], List...

E                                ^
```

intgr:
Ah sorry, the last change was a bit rushed, fixed now.

sobolevn (approved):
👍

PR #120, opened 2021-01-13
Type checking for validator callbacks

intgr:
* Restore checking of validator callbacks. Partial support for that existed until #94.
* Validator instances that have `requires_context = True` use a two-argument convention, a new `ContextValidator` protocol is created for that case.
* Changed `__call__` of builtin validators to return `None`. `NoReturn` should only be used when the function **always** raises error, but validator callbacks only raise sometimes.


rest_framework-stubs/validators.pyi
```diff
@@ -1,10 +1,17 @@
...
 from rest_framework.serializers import BaseSerializer
 
 _T = TypeVar("_T", bound=Model)
+_V = TypeVar("_V", contravariant=True)
+
+class ContextValidator(Protocol[_V]):
+    requires_context: bool
+    def __call__(self, __value: _V, __context: Field) -> None: ...
+
+Validator = Union[Callable[[_V], None], ContextValidator[_V]]
```
intgr, line 14: `ContextValidator` and `Validator` are symbols that don't exist in DRF code proper. Should I start them with underscore or keep as is?

There already are a few examples of stubs-only classes without underscore, such as `AsView`, `ViewSetAction`, `UsesQuerySet`.

sobolevn: Are these items supposed to be used by the end user? Or only by our own code? 

If they are public, then we would have to make them public and also provide some kind of a runtime fallback.

intgr: I don't expect end-users to need to reference them, unless they implement their own code that accepts validators.

For AsView, ViewSetAction I've used this trick:
```python
if TYPE_CHECKING:
    # Only defined in djangorestframework-stubs
    from rest_framework.decorators import ViewSetAction
    from rest_framework.views import AsView

class MyViewSet(GenericViewSet):
    @classonlymethod
    def as_view(cls, actions: "Optional[Dict[str, Union[str, ViewSetAction]]]" = None, **kwargs: Any) -> "AsView":
```

By importing when `TYPE_CHECKING` is set and quoting them, it doesn't break runtime.

sobolevn:
Thanks a lot!

PR #121, opened 2021-01-19
ChoiceField.choices is Iterable

mkurnikov:


sobolevn (approved):

asottile-sentry:
we've chosen in sentry to put this back to Sequence so that the choices render in a deterministic order (since `Iterable` allows `set` / `frozenset` / etc. which end up producing an indeterminate ordering)

Issue #124, opened 2021-01-27
error: Module 'rest_framework.compat' has no attribute 'coreapi'


vhtkrk:
After cloning the master branch and setting up a vanilla development environment, `pytest` fails every test with unexpected output from import errors.

This is a sample from the first test:

```
/Users/jh/drfstubs/djangorestframework-stubs/tests/typecheck/test_all_imports.yml:2:
E   pytest_mypy_plugins.utils.TypecheckAssertionError: Output is not expected:
E   Expected:
E   Actual:
E     ../../../../../../Users/jh/drfstubs/djangorestframework-stubs/rest_framework-stubs/schemas/coreapi.pyi:6: error: Module 'rest_framework.compat' has no attribute 'coreapi' (diff)
E     ../../../../../../Users/jh/drfstubs/djangorestframework-stubs/rest_framework-stubs/schemas/generators.pyi:4: error: Module 'rest_framework.compat' has no attribute 'coreapi' (diff)
```

Every test has these same two extra lines in the output.

This is somewhat puzzling because if I run `python3` and in the REPL type `from rest_framework.compat import coreapi` it imports it just fine.

I'm using Python 3.8.6 on macOS 10.14.6.
Installed packages:
```
appdirs==1.4.4
asgiref==3.3.1
attrs==20.3.0
certifi==2020.12.5
cfgv==3.2.0
chardet==4.0.0
coreapi==2.3.3
coreschema==0.0.4
decorator==4.4.2
distlib==0.3.1
Django==3.1.5
django-stubs==1.7.0
djangorestframework==3.12.1
-e git+git@github.com:vhtkrk/djangorestframework-stubs.git@746c7b7043dff53b316fee957f3bf1ad0917d2cd#egg=djangorestframework_stubs
filelock==3.0.12
gitdb==4.0.5
GitPython==3.1.9
identify==1.5.13
idna==2.10
iniconfig==1.1.1
itypes==1.2.0
Jinja2==2.11.2
MarkupSafe==1.1.1
mypy==0.800
mypy-extensions==0.4.3
nodeenv==1.5.0
packaging==20.8
pluggy==0.13.1
pre-commit==2.7.1
py==1.10.0
pyparsing==2.4.7
pystache==0.5.4
pytest==6.1.1
pytest-mypy-plugins==1.6.1
pytz==2020.5
PyYAML==5.4.1
requests==2.25.1
six==1.15.0
smmap==3.0.5
sqlparse==0.4.1
[...]

PR #125, opened 2021-01-27
Fix status.HTTP_200_OK

### LINKED ISSUE #123: status.HTTP_200_OK is Literal[202]
`rest_framework-stubs/status.pyi` defines HTTP_200_OK as follows:

    HTTP_200_OK: Literal[202] 

I believe this should be `Literal[200]`. DRF's `rest_framework/status.py` agrees:

    HTTP_200_OK = 200

vhtkrk:
`status.HTTP_200_OK` should be `Literal[200]`, not `Literal[202]`. Fixes #123.

I cannot run `pytest` because of #124, but I don't see how this could break anything.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,5 +1,4 @@
 wheel
-mypy==0.790
 gitpython==3.1.9
 pre-commit==2.7.1
 pytest==6.1.1
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,5 @@
 wheel
+mypy==0.790
 gitpython==3.1.9
 pre-commit==2.7.1
 pytest==6.1.1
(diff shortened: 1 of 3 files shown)


sobolevn:
This happens because `mypy@0.800` changes how it treats re-exports.
Right now it treats re-exports as "explicit" only if the import/export names do match.

Right: `from some import coreapi as coreapi`
Wrong: `from some import other as coreapi`

vhtkrk:
You're right; seems like pinning mypy to 0.790 fixes the tests (and then all tests pass with this change). Seems like the CI is also affected by that...

sobolevn:
@vhtkrk can you please also pin it for now in this PR? Thanks!

vhtkrk:
Done. Now also fixes #124.

sobolevn:
I guess that it is better to add it here: https://github.com/typeddjango/djangorestframework-stubs/blob/master/setup.py#L22

sobolevn:
`mypy>=0.790,<0.800`

sobolevn (approved):
Thanks!

PR #127, opened 2021-01-27
Fix model Meta type

### LINKED ISSUE #126: ModelSerializer.Meta.model should be a Type[_MT] not _MT
As it's not a model instance, but a type. Similar also to BaseManager.model in https://github.com/typeddjango/django-stubs/blob/master/django-stubs/db/models/manager.pyi#L14

palfrey:
Fixes #126. Had to cherry-pick from https://github.com/typeddjango/djangorestframework-stubs/pull/125 to make the build work.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,5 +1,4 @@
 wheel
-mypy==0.790
 gitpython==3.1.9
 pre-commit==2.7.1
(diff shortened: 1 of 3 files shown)


sobolevn (approved):
Thanks a lot!

PR #130, opened 2021-01-30
Update mypy to 0.800 and fix compatibility

intgr:
In response to discussion in #101, this PR updates mypy to 0.800 and fixes the test suite.

### DIFF
--- a/rest_framework-stubs/compat.pyi
+++ b/rest_framework-stubs/compat.pyi
@@ -52,3 +52,26 @@ def distinct(queryset: QuerySet, base: Optional[QuerySet]) -> QuerySet: ...
 SHORT_SEPARATORS: Tuple[str, str]
 LONG_SEPARATORS: Tuple[str, str]
 INDENT_SEPARATORS: Tuple[str, str]
+
+__all__ = [
+    "coreapi",
+    "coreschema",
+    "requests",
+    "postgres_fields",
+    "QuerySet",
+    "uritemplate",
+    "yaml",
+    "pygments",
+    "markdown",
+    "apply_markdown",
+    "Preprocessor",
+    "CodeBlockPreprocessor",
+    "pygments_css",
+    "pygments_highlight",
+    "md_filter_add_syntax_highlight",
+    "unicode_http_header",
+    "distinct",
+    "SHORT_SEPARATORS",
+    "LONG_SEPARATORS",
+    "INDENT_SEPARATORS",
+]
--- a/scripts/typecheck_tests.py
+++ b/scripts/typecheck_tests.py
@@ -70,7 +70,6 @@ IGNORED_ERRORS = {
         'List item 0 has incompatible type "Type[',
         'error: Module has no attribute "coreapi"',
         'Value of type "Optional[str]" is not indexable',
-        "Module 'rest_framework.compat' has no attribute 'yaml'",
         'Incompatible types in assignment (expression has type "AsView[GenericView]", variable has type "AsView[Callable[[Any], Any]]")',  # noqa: E501
     ],
     "browsable_api": [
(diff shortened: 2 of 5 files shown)


intgr:
I'm still confused why it didn't fail before.

In https://github.com/typeddjango/djangorestframework-stubs/pull/101#issuecomment-770180501 @sobolevn said that

> mypy 0.800 introduced new explicit re-export rule.
> 
> Now:
> 
> * This is an explicit re-export: `from a import b as b`
> * This is not: `from a import b as c`
> 
> And we need to make sure that we meet this new rule.

However, the errors are about `rest_framework.compat` module, which **never** actually contained re-exports, `compat.pyi` had code:

```python
import coreapi  # noqa: F401
try:
    import coreschema
except ImportError:
    coreschema = None  # type: ignore
```
etc. These shouldn't be re-exports under the old rules either, right?  Then why didn't these errors show up before?

```
drf_source/tests/schemas/test_coreapi.py:12: error: Module 'rest_framework.compat' has no attribute 'coreapi'
drf_source/tests/schemas/test_coreapi.py:12: error: Module 'rest_framework.compat' has no attribute 'coreschema'
drf_source/tests/schemas/test_managementcommand.py:11: error: Module 'rest_framework.compat' has no attribute 'uritemplate'
drf_source/tests/schemas/test_openapi.py:11: error: Module 'rest_framework.compat' has no attribute 'uritemplate'
drf_source/tests/test_api_client.py:9: error: Module 'rest_framework.compat' has no attribute 'coreapi'
drf_source/tests/test_api_client.py:9: error: Module 'rest_framework.compat' has no attribute 'coreschema'
drf_source/tests/test_encoders.py:9: error: Module 'rest_framework.compat' has no attribute 'coreapi'
drf_source/tests/test_fields.py:16: error: Module 'rest_framework.fields' has no attribute 'DjangoImageField'; maybe "ImageField"?
drf_source/tests/test_filters.py:13: error: Module 'rest_framework.compat' has no attribute 'coreschema'
drf_source/tests/test_renderers.py:16: error: Module 'rest_framework.compat' has no attribute 'coreapi'
drf_source/tests/test_templatetags.py:7: error: Module 'rest_framework.compat' has no attribute 'coreapi'
drf_source/tests/test_templatetags.py:7: error: Module 'rest_framework.compat' has no attribute 'coreschema'
```

rest_framework-stubs/compat.pyi
```diff
@@ -52,3 +52,26 @@ def distinct(queryset: QuerySet, base: Optional[QuerySet]) -> QuerySet: ...
 SHORT_SEPARATORS: Tuple[str, str]
 LONG_SEPARATORS: Tuple[str, str]
 INDENT_SEPARATORS: Tuple[str, str]
+
+__all__ = [
+    "coreapi",
+    "coreschema",
+    "requests",
+    "postgres_fields",
```
intgr, line 60: Since `from django.contrib.postgres import fields as postgres_fields` isn't treated as a re-export, it seems the only way to re-export it is via `__all__`?

scripts/typecheck_tests.py
```diff
@@ -70,7 +70,6 @@
         'List item 0 has incompatible type "Type[',
         'error: Module has no attribute "coreapi"',
         'Value of type "Optional[str]" is not indexable',
-        "Module 'rest_framework.compat' has no attribute 'yaml'",
```
intgr, line 73: So this did cause errors for `rest_framework.compat.yaml`.

Why didn't it cause errors for `rest_framework.compat.coreapi` etc?  🤔

setup.py
```diff
@@ -20,7 +20,7 @@ def find_stub_files(name):
     readme = f.read()
 
 dependencies = [
-    "mypy>=0.790,<0.800",
+    "mypy>=0.800",
```
sobolevn, line 23: Can we use `>=0.790`?

rest_framework-stubs/compat.pyi
sobolevn: Looks like it, can we please test it with `tests/typecheck`?

sobolevn (approved):
Awesome! Thanks!

PR #131, opened 2021-02-01
Make some mixin methods generic

henribru:
This change makes certain mixin methods generic using a protocol for the self type. This ensures that the queryset and instance arguments for the methods have specific types based on the model instead of `BaseSerializer[Any]` and `Model`. For `CreateModelMixin` and `UpdateModelMixin` this isn't a huge deal, because you were already allowed to override `BaseSerializer` with e.g. `BaseSerializer[MyModel]` (the tests I've included for those actually pass even without the change). It can probably still be useful in some niche cases though. For `DestroyModelMixin` it's a big win since overriding `Model` with `MyModel` would produce an error without this change.


sobolevn (approved):
Awesome! Thanks!

rest_framework-stubs/mixins.pyi
```diff
@@ -1,13 +1,17 @@
...
 from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework.serializers import BaseSerializer
 
+_MT = TypeVar("_MT", bound=Model)
+
 class CreateModelMixin:
     def create(self, request: Request, *args: Any, **kwargs: Any) -> Response: ...
-    def perform_create(self, serializer: BaseSerializer) -> None: ...
+    def perform_create(self: UsesQuerySet[_MT], serializer: BaseSerializer[_MT]) -> None: ...
```
intgr, line 14: Wow, I didn't know you could do that 😮 nice.

antonagestam: @intgr mypy has a section on ["Advanced self-types"](https://mypy.readthedocs.io/en/stable/more_types.html#advanced-uses-of-self-types). Pretty cool feature! 👍

PR #133, opened 2021-03-10
Make the APIView Mixin Model type covariant

### LINKED ISSUE #29: Argument 1 of "perform_create" is incompatible with supertype "CreateModelMixin"; supertype defines the argument type as "BaseSerializer"
Code:

```python
from rest_framework import viewsets

from server.apps.main.logic import repo
from server.apps.main.serializers import BlogPostSerializer
from server.apps.main.models import BlogPost


class BlogPostViewset(viewsets.ModelViewSet):
    """API demo."""

    serializer_class = BlogPostSerializer
    queryset = repo.published_posts()

    def perform_create(self, serializer: BlogPostSerializer) -> None:
        serializer.save()
```

Output:

```python
» PYTHONPATH="$PYTHONPATH:$PWD" mypy server
server/apps/main/views.py:28: error: Argument 1 of "perform_create" is incompatible with supertype "CreateModelMixin"; supertype defines the argument type as "BaseSerializer"
```

I guess that variance of the argument is incorrect.

w0rp:
This should fix #29 for mixins, which currently doesn't let you use a `Model` subtype for method overrides on `perform_create` and other methods.

If anyone can point me in the direction of where to add tests, let me know and I'll add some.


sobolevn:
Looks like we have a problem: `typeddjango/djangorestframework-stubs/rest_framework-stubs/mixins.pyi:30: error: Cannot use a covariant type variable as a parameter`

What do you think on this one?

w0rp:
I can fix that. It will mean we'll need to have the mixins inherit `UsesQuerySet` for everything to work. I think it's kind of an oversight in Python typing, because it really ought to be able to figure it out because of the way `self` is typed for the methods.

w0rp:
I tried a few different things, and my conclusion right now is that Python's current type system makes modelling this kind of thing without adding comments to ignore type errors each time you write an override impossible. There are a lot of restrictions put on methods to prevent you hypothetically adding any Animal to a list of Dogs or something, but all of the errors prevent you from writing code for real applications like DRF without the type checker telling you off.

PR #139, opened 2021-04-01
Make MultipleChoiceField initial and default args Set

### LINKED ISSUE #138: MultipleChoiceField should take default/initial value as a 'Set'
__Repro:__

```python
from rest_framework.fields import MultipleChoiceField
MultipleChoiceField(choices=[(1, "1"), (2, "2")], default={1})
```

Raises:
```
error: Argument "default" to "MultipleChoiceField" has incompatible type "Set[int]"; expected "Union[_Empty, Sequence[str], Sequence[int], Callable[[], Sequence[str]], Callable[[], Sequence[int]]]"  [arg-type
```

[DRF docs for MultipleChoiceField](https://www.django-rest-framework.org/api-guide/fields/#multiplechoicefield) states (in the end of first paragraph):
> .. `to_internal_value` returns a `set` containing the selected values.

Which means that setting for instance: `default=[1]` and when the _default is triggered_ it'll return a `list[int]` type. Though when input is provided to the `MultipleChoiceField` (and `to_internal_value` is triggered) it'll return a `set[int]`.

---

Same story goes for `MultipleChoiceField.to_representation` and `MultipleChoiceField.initial`.

flaeppe:
Fixes #138

### DIFF
--- a/rest_framework-stubs/fields.pyi
+++ b/rest_framework-stubs/fields.pyi
@@ -515,10 +515,10 @@ class MultipleChoiceField(
         write_only: bool = ...,
         required: bool = ...,
         default: Union[
-            _Empty, Sequence[str], Sequence[int], Callable[[], Sequence[str]], Callable[[], Sequence[int]]
+            _Empty, Set[str], Set[int], Callable[[], Set[str]], Callable[[], Set[int]]
         ] = ...,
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Thanks!

PR #140, opened 2021-04-08
fix: use Type annotations for ModelSerializer and HyperlinkedModelSerializer `serializer_XXX` attributes

swistakm:
Following fields that can be configured in `ModelSerializer` should be types (field classes) instead of fields instances:
- `serializer_related_field`
- `serializer_related_to_field`
- `serializer_url_field`
- `serializer_choice_field`

I have included test that roughly redefines the same attribute values as [rest_framework.serializers.ModelSerializer](https://github.com/encode/django-rest-framework/blob/fd017d00f938c6629d2eeb1b4d81716dff6d006e/rest_framework/serializers.py#L889-L892):

```python
from rest_framework import serializers
from django.contrib.auth.models import User

class TestSerializer(serializers.ModelSerializer):
    serializer_related_field = serializers.PrimaryKeyRelatedField
    serializer_related_to_field = serializers.SlugRelatedField
    serializer_url_field = serializers.HyperlinkedIdentityField
    serializer_choice_field = serializers.ChoiceField

    class Meta:
        model = User
```

The same typing rule applies to following field of `HyperlinkedModelSerializer`:
- `serializer_related_field`

And following is example serializer class based on implementation of [rest_framework.serializers.HyperlinkedModelSerializer](https://github.com/encode/django-rest-framework/blob/fd017d00f938c6629d2eeb1b4d81716dff6d006e/rest_framework/serializers.py#L1617):

```python
from rest_framework import serializers
from django.contrib.auth.models import User

class TestSerializer(serializers.HyperlinkedModelSerializer):
    serializer_related_field = serializers.HyperlinkedRelatedField

    class Meta:
        model = User
 ```

This base serializer class inherits from `ModelSerializer` so it can reuse parent typing annotations. I have added additional test that covers `HyperlinkedModelSerializer` to avoid future regression.


sobolevn (approved):
Awesome! Thanks!

PR #141, opened 2021-04-10
fix: APIView.dispatch expects a HttpRequest which gets transformed into a Request by calling initialize_request

LanDinh:
Current, broken behavior: `dispatch` and `initialize_request` both expect `Request` as argument type.
Actual flow: `dispatch` receives a `HttpRequest`, then proceeds to calling `initialize_request` in order to transform the `HttpRequest` into a `Request`, before doing anything else. So both `dispatch` and `initialize_request` should expect a `HttpRequest` as argument instead of a `Request`.

Incidentally, this means that a `# type: ignore[override]` can be deleted as well.


LanDinh:
Seems like Travis cancelled the build because the project ran out of credits...
I know that django-stubs has already migrated to Github actions, I guess that this is a possible plan of action for this project, too?

rest_framework-stubs/views.pyi
```diff
@@ -81,10 +81,10 @@ class APIView(View):
...
         self, request: Request, *args: Any, **kwargs: Any
     ) -> Tuple[Optional[str], Optional[BaseVersioning]]: ...
-    def initialize_request(self, request: Request, *args: Any, **kwargs: Any) -> Request: ...
+    def initialize_request(self, request: HttpRequest, *args: Any, **kwargs: Any) -> Request: ...
     def initial(self, request: Request, *args: Any, **kwargs: Any) -> None: ...
     def finalize_response(self, request: Request, response: Response, *args: Any, **kwargs: Any) -> Response: ...
     def handle_exception(self, exc: Exception) -> Response: ...
     def raise_uncaught_exception(self, exc: Exception) -> NoReturn: ...
-    def dispatch(self, request: Request, *args: Any, **kwargs: Any): ...  # type: ignore[override]
+    def dispatch(self, request: HttpRequest, *args: Any, **kwargs: Any): ...
```
sobolevn, line 89: What do we return here?

tests/typecheck/test_views.yml
```diff
@@ -19,3 +19,14 @@
         queryset = MyModel.objects.all()
 
         def perform_destroy(self, instance: MyModel) -> None: ...
+
+- case: test_build_api_request
+  main: |
+    from django.http.request import HttpRequest
+    from rest_framework.views import APIView
+
+    request = HttpRequest()
```
sobolevn, line 28: It is better to write: `request: HttpRequest`

sobolevn (commented):
Thanks!

rest_framework-stubs/views.pyi
LanDinh: Adjusted to `HttpResponseBase` from what I could see in the implementation of it (which also made me realize that this is wrong in `django-stubs` - it types the return value for `dispatch` as `HttpResponse`, which prevents `StreamingHttpResponse` from being a valid value...

tests/typecheck/test_views.yml
LanDinh: Done

rest_framework-stubs/views.pyi
```diff
@@ -86,5 +86,5 @@ class APIView(View):
     def finalize_response(self, request: Request, response: Response, *args: Any, **kwargs: Any) -> Response: ...
     def handle_exception(self, exc: Exception) -> Response: ...
     def raise_uncaught_exception(self, exc: Exception) -> NoReturn: ...
-    def dispatch(self, request: HttpRequest, *args: Any, **kwargs: Any): ...
+    def dispatch(self, request: HttpRequest, *args: Any, **kwargs: Any): HttpResponseBase
```
sobolevn, line 89: Typo!

```suggestion
    def dispatch(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponseBase:
```

sobolevn (changes requested):

sobolevn:
> I know that django-stubs has already migrated to Github actions, I guess that this is a possible plan of action for this project, too?

Yes, please! 
Can you please copy the setup from `django-stubs`?

LanDinh (commented):
Thanks for those fixes!

LanDinh:
Now I'm confused by the Github interface..... It has this big fat red thing of "changes requested". When I click on "show changes", it shows me the things that I marked as reviewed in my last comment and are thus already in the PR. So what is this thing that Github thinks requires changes??

sobolevn (approved):
Thanks!

PR #142, opened 2021-05-01
Small improvements to rest internals

LanDinh:
1. Why `force_authenticate` should expect a `HttpRequest` instead of a `Request`:
    When looking at `request.py`, this code snipped stands out:
    ```
    class Request::
      ...
      def __init__(self, request, parsers=None, authenticators=None, negotiator=None, parser_context=None):
        ...
        force_user = getattr(request, '_force_auth_user', None)
        force_token = getattr(request, '_force_auth_token', None)
        if force_user is not None or force_token is not None:
          forced_auth = ForcedAuthentication(force_user, force_token)
          self.authenticators = (forced_auth,)
        ...
    ```
    Notably: the parameter `request` is of type `HttpRequest`. Meaning that `force_authentication` sets the attributes `_force_auth_user` and `_force_auth_token` on the `HttpRequest`, before the `Request` has even be created.

2. Missing attribute `csrf_exempt` on the return value of `as_view`:
    `APIView.as_view` applies the method decorator `csrf_exempt` on the view, which ends up setting the attribute `csrf_exempt = True` on the view function. I wish it would be possible to annotate this in django directly, but unfortunately, that doesn't work, because this attribute only gets set when applying this decorator, and is thus not present when it doesn't get applied - and I haven't been able to figure out how to annotate an attribute which might not be present on an object (note the difference between setting an attribute to `None` and the attribute not existing at all, which is mainly the difference in the return value of `hasattr`). But since DRF applies this decorator to all views using the `as_view` function, it should be okay to add this type hint. This shouldn't affect the Django `View`, since that already has a different type annotation (a simple `Callable` without any of the already-annotated properties)


sobolevn:
Yey! It fails!

LanDinh:
I'll wait to take another look at this until the general code base is green again :)

sobolevn:
The `master` branch is green, please, feel free to continue! 👍

LanDinh:
So as expected, this PR doesn't break any of the tests :)

sobolevn (approved):
Thanks!

PR #143, opened 2021-05-01
Migrate to GitHub actions

LanDinh:
I've copied over the setup from django-stubs and adapted it to what I could see in the .travis.yml.

Notably:

- pytest is only run on Python 3.7 for django-stubs, but on multiple Python versions for djangorestframework-stubs
- Instead, djangorestframework-stubs doesn't test against multiple Django / DRF versions

But of course I might have been misreading things. Let's see if the tests pass!

FYI: This also includes the bug & PR templates from django-stubs.


LanDinh:
Ups. Based it on the wrong branch, lemme fix that.

sobolevn (approved):
Not sure, why it was not executed. Let's try to merge it. 

Thanks a lot! 👍

sobolevn:
Action: https://github.com/typeddjango/djangorestframework-stubs/actions/runs/802038426

LanDinh:
Funnily enough, after looking through these failures: it was one of my previous PRs in which I declared the return type `HttpResponseBase` for `ApiView.dispatch`. Which means that as soon as typeddjango/django-stubs#607 is merged, rerunning the tests should turn them green again :D

I guess this is what happens without CI testing :/ "was working fine with my project, so of course it has to be right!"

LanDinh:
Actually, I just realized that the tests don't run against the django-stubs master, but against the latest released verison, right? In that case, lemme add some `# type: ignore` hints to bridge the time till then

LanDinh:
Done with #144

PR #144, opened 2021-05-01
Fix tests now that they are operational again

LanDinh:
# I have made things!

Seeing as djangorestframework-stubs ran out of Travis credits and has thus been without operational CI tests for some time, some errors sneaked in. Now that the migration to github actions is done, fix these failing tests.


sobolevn (approved):
Thanks!

PR #146, opened 2021-05-25
Make APIView.versioning_class type, not str

antonagestam:
Hi 👋

This changes `APIView.versioning_class` to `type` instead of `str`. See [docs here](https://www.django-rest-framework.org/api-guide/versioning/#configuring-the-versioning-scheme) and [here's some code assuming it's a callable](https://github.com/encode/django-rest-framework/blob/master/rest_framework/views.py#L378).


sobolevn (approved):

sobolevn:
Thanks!

PR #151, opened 2021-06-18
Add Mypy 0.900 compatibility

henribru:
- Adds a dependency on a couple stub packages that Mypy suggested since it doesn't bundle typeshed anymore in 0.900 (added `types-pytz` as a dev-only dependency because `pytz` is never imported in the stubs, but it's used in DRF's test suite)
- Fixes the test output to match Mypy 0.900


sobolevn (approved):
Great work! Thanks a lot!

PR #152, opened 2021-06-22
Make remaining fields accept default=None/initial=None

intgr:
ChoiceField, MultipleChoiceField, FilePathField, FileField, ImageField, ListField, DictField, HStoreField, JSONField were previously using hard-coded `default` and `initial` constructor argument types.

Adapted the `_DefaultInitial[]` type alias instead -- introduced in PR #101 -- which allows `None`.


tests/typecheck/test_fields.yml
```diff
@@ -78,15 +78,86 @@
         x: Optional[str] = CharField().get_initial()
         y: Optional[int] = CharField().get_initial()  # E: Incompatible types in assignment (expression has type "Optional[str]", variable has type "Optional[int]")
 
-        MultipleChoiceField(choices=[(1, "1"), (2, "2")], default={1})
-        MultipleChoiceField(choices=[(1, "1"), (2, "2")], default=[1])  # E: Argument "default" to "MultipleChoiceField" has incompatible type "List[int]"; expected "Union[_Empty, Set[str], Set[int], Callable[[], Set[str]], Callable[[], Set[int]]]"
-
-        MultipleChoiceField(choices=[(1, "1"), (2, "2")], initial={1})
-        MultipleChoiceField(choices=[(1, "1"), (2, "2")], initial=[1])  # E: Argument "initial" to "MultipleChoiceField" has incompatible type "List[int]"; expected "Union[_Empty, Set[str], Set[int], Callable[[], Set[str]], Callable[[], Set[int]]]"
```
intgr, line 85: I moved these tests to the separate `MultipleChoiceField_default` test case.

rest_framework-stubs/fields.pyi
```diff
@@ -514,12 +514,8 @@ class MultipleChoiceField(
...
         write_only: bool = ...,
         required: bool = ...,
-        default: Union[
-            _Empty, Set[str], Set[int], Callable[[], Set[str]], Callable[[], Set[int]]
-        ] = ...,
-        initial: Union[
-            _Empty, Set[str], Set[int], Callable[[], Set[str]], Callable[[], Set[int]]
-        ] = ...,
+        default: _DefaultInitial[Union[Set[Union[str, int]], Set[str], Set[int]]] = ...,
+        initial: _DefaultInitial[Union[Set[Union[str, int]], Set[str], Set[int]]] = ...,
```
intgr, line 518: Since `Set` is mutable, `Set[str]` is not compatible with `Set[Union[str, int]]`. So I explicitly allowed all three combinations:
* `Set[Union[str, int]]`
* `Set[str]`
* `Set[int]`

sobolevn (approved):
Great work, thank you!

PR #153, opened 2021-06-25
fix ModelSerializer.serializer_field_mapping value type

etene:
I'm pretty sure the annotation for the values [`ModelSerializer.serializer_field_mapping`](https://github.com/encode/django-rest-framework/blob/71e6c30034a1dd35a39ca74f86c371713e762c79/rest_framework/serializers.py#L855) is missing a `Type`, as it doesn't expect field instances but field classes instead.

This commit fixes the problem for me, I hope it's correct !

## Related issues

None, sorry this is a very small fix


sobolevn (approved):
Yes, you are correct: https://github.com/encode/django-rest-framework/blob/71e6c30034a1dd35a39ca74f86c371713e762c79/rest_framework/serializers.py#L855-L882

Thanks!

PR #154, opened 2021-06-27
Adds DjangoFilterBackend option for filter_backends

### LINKED ISSUE #150: DjangoFilterBackend for filter_backends
# Bug report

## What's wrong

[`GenericAPIView`](https://github.com/typeddjango/djangorestframework-stubs/blob/master/rest_framework-stubs/generics.pyi ) defines `filter_backends` as `Sequence[Type[BaseFilterBackend]]`. However, the [DRF examples](https://www.django-rest-framework.org/api-guide/filtering/#setting-filter-backends) include the code sample `filter_backends = [django_filters.rest_framework.DjangoFilterBackend]` and `DjangoFilterBackend` does not extend `BaseFilterBackend`

## How is that should be

`DjangoFilterBackend` should be allowed.

## System information

- OS:
- `python` version: 3.7
- `django` version: 3.2.4 
- `mypy` version: 0.812
- `django-stubs` version: 1.8.0

palfrey:
## Related issues

Closes #150


rest_framework-stubs/generics.pyi
```diff
@@ -9,6 +9,7 @@ from rest_framework.pagination import BasePagination
 from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework.serializers import BaseSerializer
+from django_filters.rest_framework import DjangoFilterBackend
```
sobolevn, line 12: Sorry, we cannot do that. Because we can't rely on third party projects like `django-filters` here.

sobolevn (commented):
I see several solutions here:
1. If this is a common problem, we add a protocol version of `BaseFilterBackend`
2. Or we can add this example to docs and suggest to use `type: ignore` there

What do you think?

palfrey:
> I see several solutions here:
>     1. If this is a common problem, we add a protocol version of `BaseFilterBackend`

AFAIK, the only common case is this one! OTOH, it's used enough. Given the need for not having third-party deps, I can make a protocol version.

rest_framework-stubs/generics.pyi
```diff
@@ -19,12 +18,19 @@ def get_object_or_404(
...
+    def get_schema_fields(self, view: views.APIView) -> List[Any]: ...
+    def get_schema_operation_parameters(self, view: views.APIView): ...
+
 class GenericAPIView(views.APIView, UsesQuerySet[_MT_co]):
     queryset: Optional[Union[QuerySet[_MT_co], Manager[_MT_co]]] = ...
     serializer_class: Optional[Type[BaseSerializer]] = ...
     lookup_field: str = ...
     lookup_url_kwarg: Optional[str] = ...
-    filter_backends: Sequence[Type[BaseFilterBackend]] = ...
+    filter_backends: Sequence[Type[BaseFilterProtocol]] = ...
```
sobolevn, line 33: Maybe a `Union[]`? It would be just more readable.

And please, add a test with something like `django-filter` has! 👍

sobolevn (commented):
Great!

rest_framework-stubs/generics.pyi
palfrey: Can the test use `django-filter` even if the main package doesn't have a dep on it?

intgr:
Maybe we can even ensure that the model type is compatible...

```python
class BaseFilterProtocol(Protocol[_MT_co]):
     def filter_queryset(self, request: Request, queryset: QuerySet[_MT_co], view: views.APIView) -> QuerySet[_MT_co]: ...
```

```python
    filter_backends: Sequence[Type[Union[BaseFilterBackend, BaseFilterProtocol[_MT_co]]]] = ...
```

rest_framework-stubs/generics.pyi
sobolevn: We can create a custom class with the same signature.

sobolevn:
@intgr what do you think about adding the same to `BaseFilterBackend`?

intgr:
Yes, would be great, though not necessarily in this particular PR.

palfrey:
> Maybe we can even ensure that the model type is compatible...
> 
> ```python
> class BaseFilterProtocol(Protocol[_MT_co]):
>      def filter_queryset(self, request: Request, queryset: QuerySet[_MT_co], view: views.APIView) -> QuerySet[_MT_co]: ...
> ```
> 
> ```python
>     filter_backends: Sequence[Type[Union[BaseFilterBackend, BaseFilterProtocol[_MT_co]]]] = ...
> ```

2c67a0d01b797ae747204e28d795e6b8eef20973 does this, but I keep getting the informative error ` error: Covariant type variable "_MT_co" used in protocol where invariant one is expected (diff)` that I'm not sure how to solve. Any ideas?

rest_framework-stubs/generics.pyi
palfrey: I've added a test for this, but it's failing for other reasons

palfrey:
@intgr @sobolevn Any thoughts on the test issues here?

rest_framework-stubs/generics.pyi
```diff
@@ -19,12 +19,19 @@ def get_object_or_404(
 class UsesQuerySet(Protocol[_MT_co]):
     def get_queryset(self) -> QuerySet[_MT_co]: ...
 
+# Can't just use BaseFilterBackend because there's also things like django_filters.rest_framework.DjangoFilterBackend that are
+# valid options but don't extend it
+class BaseFilterProtocol(Protocol[_MT_co]):
```
sobolevn, line 24: Maybe you should create a new type var to please the type tests? Check out logs to see the suggestions.

palfrey: Works now :)

sobolevn (approved):
Thanks! This a really useful special case! 👍

palfrey:
Any chance of a new release with this in?

palfrey:
> Any chance of a new release with this in?

Ping? @sobolevn?

palfrey:
Got released in [1.5.0](https://github.com/typeddjango/djangorestframework-stubs/releases/tag/1.5.0)

PR #155, opened 2021-07-11
Make APIView.content_negotiation_class type, not str

flaeppe:
Hi,

This changes `APIView.content_negotiation_class` to `type` instead of `Optional[str]`. See [docs on content negotiation](https://www.django-rest-framework.org/api-guide/content-negotiation/#setting-the-content-negotiation) and [some code assuming it's a callable](https://github.com/encode/django-rest-framework/blob/98e56e0327596db352b35fa3b3dc8355dc9bd030/rest_framework/views.py#L291).


sobolevn (approved):
Thanks!

PR #156, opened 2021-07-11
Add end-of-file-fixer hook to pre-commit

flaeppe:
Not sure if this hook is something you wish to use, though I just happened to found a couple of files missing new line or ending with an empty one. Thought I might just add it in in case you wanted it.


sobolevn (approved):

sobolevn:
You can add the same here https://github.com/typeddjango/django-stubs/blob/master/.pre-commit-config.yaml

flaeppe:
@sobolevn Cool, I opened https://github.com/typeddjango/django-stubs/pull/668

PR #161, opened 2021-07-24
Serializer iterator returns BoundField, not str.

danielroseman:
See https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L529


sobolevn (approved):
Thanks!

PR #162, opened 2021-08-03
Add permissions.OperandHolder to decorators.permission_classes signature

kontsaki:
Hello, this will allow `permission_classes` to accept types like `BasePermission & BasePermission`.
i am not sure if i should import `_PermissionsClasses` as defined in the permissions stub and remove the preceding underscore, feedback is appreciated.

### DIFF
--- a/tests/typecheck/test_decorators.yml
+++ b/tests/typecheck/test_decorators.yml
@@ -24,3 +24,8 @@
         from rest_framework.decorators import api_view
         @api_view()  # E: Value of type variable "_View" of function cannot be "Callable[[Any], List[Any]]"
         def view_func2(request) -> list: ...
+
+-   case: permission_classes
+    main: |
+        from rest_framework.decorators import permission_classes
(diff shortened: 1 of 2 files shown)


sobolevn (changes requested):
Hi, thanks! Let's add for this!

kontsaki:
oh yes, on it.

kontsaki:
would a test like this suffice ?

tests/typecheck/test_decorators.yml
```diff
@@ -24,3 +24,8 @@
         from rest_framework.decorators import api_view
         @api_view()  # E: Value of type variable "_View" of function cannot be "Callable[[Any], List[Any]]"
         def view_func2(request) -> list: ...
+
+-   case: permission_classes
+    main: |
+        from rest_framework.decorators import permission_classes
```
sobolevn, line 30: We need to test that `&` form works. You can also add tests for other ways if
- they are missing
- you have the time for it 🙂

kontsaki:
i can add more tests if you have something else in mind

sobolevn (approved):
Thanks!

PR #170, opened 2021-11-03
Allow fields to be None

greg-finley:
`fields` could be `None` if `exclude` is provided. Though I guess no way to represent in types that you need one or the other populated.

### DIFF
--- a/rest_framework-stubs/serializers.pyi
+++ b/rest_framework-stubs/serializers.pyi
@@ -221,7 +221,7 @@ class ModelSerializer(Serializer, BaseSerializer[_MT]):
     instance: Optional[Union[_MT, Sequence[_MT]]]  # type: ignore[override]
     class Meta:
         model: Type[_MT] # type: ignore
-        fields: Union[Sequence[str], Literal["__all__"]]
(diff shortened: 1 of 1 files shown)


sobolevn:
@greg-finley would you mind attaching a link to the DRF docs to this PR, please?

greg-finley:
https://www.django-rest-framework.org/api-guide/serializers/

sobolevn:
Reopened to trigger the CI once again.

intgr:
My understanding is that `fields` may be **omitted**. But that doesn't mean it can be set to `None`.

Per DRF documentation
```python
class AccountSerializer(serializers.ModelSerializer):
    class Meta:
        model = Account
        exclude = ['users']
```

There's no `fields = None` here.

greg-finley:
I have code similar to the below. It passes unit tests that the latter will exclude a field and return the other fields.

```
class AccountSerializer(serializers.ModelSerializer):
    class Meta:
        model = Account
        fields: Optional[Union[Sequence[str], Literal["__all__"]]] = '__all__'

class AccountHideUsersSerializer(AccountSerializer):
    class Meta(AccountSerializer.Meta):
        model = Account
        fields = None
        exclude = ['users']
```

PR #187, opened 2022-01-29
Use pre-commit action

antonagestam:
Attempt to fix pre-commit breaking on master by using the pre-commit action.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,17 +15,10 @@ jobs:
         python-version: ['3.9']
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+      - uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
-        run: |
-          pip install -U pip setuptools wheel
-          pip install -r ./requirements.txt
-
-      - name: Run pre-commit
-        run: pre-commit install && pre-commit run --all-files
+      - uses: pre-commit/action@v2.0.3
 
   test:
     runs-on: ubuntu-latest


.github/workflows/test.yml
```diff
@@ -15,17 +15,10 @@ jobs:
...
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
-        run: |
-          pip install -U pip setuptools wheel
-          pip install -r ./requirements.txt
-
-      - name: Run pre-commit
-        run: pre-commit install && pre-commit run --all-files
+      - uses: pre-commit/action@v2.0.3
```
sobolevn, line 21: I think it is deprecated 🤔 
We can use pre-commitci

antonagestam: Oh, that's sad news, I'll probably implement it as a reusable workflow in [antonagestam/python-build](https://github.com/antonagestam/python-build) because I really like the idea of everything running within Github Actions (for my own projects that is). I'll close this PR for now.

antonagestam: See https://github.com/antonagestam/phantom-types/pull/185 that now links to https://github.com/antonagestam/python-build/blob/main/.github/workflows/pre-commit.yaml for reference and let me know if you'd want to use that. I like it for the simplicity and transparency :)

sobolevn: Yes, looks good! Is it possible to send PRs with fixes back?

antonagestam: @sobolevn It might be feasible, but I suspect the same limitations would apply as for the action, e.g. [this](https://github.com/pre-commit/action#:~:text=while%20you%20could%20technically%20configure%20this%20for%20a%20public%20repository%20(using%20a%20personal%20access%20token)%2C%20I%20can%27t%20think%20of%20a%20way%20to%20do%20this%20safely%20without%20exposing%20a%20privileged%20token%20to%20pull%20requests%20%2D%2D%20if%20you%20have%20any%20ideas%2C%20please%20leave%20an%20issue!):

> while you could technically configure this for a public repository (using a personal access token), I can't think of a way to do this safely without exposing a privileged token to pull requests -- if you have any ideas, please leave an issue!

sobolevn: Ok, CI is good enough for now! 👍 
Thank you!

PR #188, opened 2022-01-30
Use reusable pre-commit workflow

antonagestam:
Supersedes #187.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,23 +9,9 @@ on:
 
 jobs:
   lint:
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: ['3.9']
-    steps:
-      - uses: actions/checkout@v2
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
-        with:
-          python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
-        run: |
-          pip install -U pip setuptools wheel
-          pip install -r ./requirements.txt
-
-      - name: Run pre-commit
-        run: pre-commit install && pre-commit run --all-files
+    uses: less-action/reusables/.github/workflows/pre-commit.yaml@v1
+    with:
+      python-version: '3.9'
 
   test:
     runs-on: ubuntu-latest


.github/workflows/test.yml
```diff
@@ -9,23 +9,9 @@ on:
...
-        with:
-          python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
-        run: |
-          pip install -U pip setuptools wheel
-          pip install -r ./requirements.txt
-
-      - name: Run pre-commit
-        run: pre-commit install && pre-commit run --all-files
+    uses: less-action/reusables/.github/workflows/pre-commit.yaml@v1
```
sobolevn, line 12: Can you please pin this to some existing commit hash?

.pre-commit-config.yaml
```diff
@@ -37,3 +37,10 @@ repos:
           - pytest
           - djangorestframework
           - types-pytz
+          - django-stubs
```
sobolevn, line 40: Do we need to pin any of these? Can it some existing definitions (like `setup.py` or `requirements.txt`)?

.github/workflows/test.yml
antonagestam: Can do! 👍

.pre-commit-config.yaml
antonagestam: There's  a lot of feature requests for something like that added to pre-commit, across at least these threads:

- https://github.com/pre-commit/mirrors-mypy/issues/3
- https://github.com/pre-commit/pre-commit/issues/730
- https://github.com/pre-commit/pre-commit/pull/880#issuecomment-446286115

I can sympathize with the maintainer not wanting to introduce something that won't work with caching in a stable way or that will be too Python-specific, but the status quo isn't very attractive tbh. Let me know if that feels like a show-stopper for going this route.

Otherwise with regards to pinning, and I don't think we need to. These only had lower limits in the requirements.txt file and I think that's what they should have here as well.

PS. in one of the linked threads Anthony mentions scripting to keep requirements.txt and `additional_dependencies` in sync. That sounds interesting to me, it would be kind of cool to have pre-commit hook that checks that `additional_dependencies` is a super-set of requirements.txt, **and** is able to fix it them if it isn't. Perhaps I'll play with that thought if I find some time 🙂

antonagestam: > Let me know if that feels like a show-stopper for going this route.

To clarify: what I mean by this is that, currently to use mirrors-mypy, we're required to accept the trade-off that we have to keep `additional_dependencies` in sync with requirements.txt.

sobolevn: Ok, let's use the naive approach! I don't think that we will have too many deps in this project. Thanks for the info! TIL

antonagestam: There were actually a lot fewer dependencies needed than I thought, I reduced it to djangorestframework and django-stubs only :)

PR #203, opened 2022-03-26
Add django-stubs v1.10.0 compatibility

### LINKED ISSUE #202: AttributeError for extract_django_settings_module when combined with newest django-stubs
# Bug report

## What's wrong
When combining djangorestframework-stubs 1.4 with the newly released django-stubs 1.10.0, Mypy will fail with:
```
Error constructing plugin instance of NewSemanalDRFPlugin

Traceback (most recent call last):
  File "/home/henribru/.cache/pypoetry/virtualenvs/lightkeeper-nCOOTSLY-py3.10/bin/mypy", line 8, in <module>
    sys.exit(console_entry())
  File "/home/henribru/.cache/pypoetry/virtualenvs/lightkeeper-nCOOTSLY-py3.10/lib/python3.10/site-packages/mypy/__main__.py", line 12, in console_entry
    main(None, sys.stdout, sys.stderr)
  File "mypy/main.py", line 96, in main
  File "mypy/main.py", line 173, in run_build
  File "mypy/build.py", line 180, in build
  File "mypy/build.py", line 231, in _build
  File "mypy/build.py", line 478, in load_plugins
  File "mypy/build.py", line 456, in load_plugins_from_config
  File "/home/henribru/.cache/pypoetry/virtualenvs/lightkeeper-nCOOTSLY-py3.10/lib/python3.10/site-packages/mypy_drf_plugin/main.py", line 25, in __init__
    django_settings_module = mypy_django_main.extract_django_settings_module(options.config_file)
AttributeError: module 'mypy_django_plugin.main' has no attribute 'extract_django_settings_module'
```

This is happening because `extract_django_settings_module` was removed in https://github.com/typeddjango/django-stubs/commit/edec5a1c99a784208d5d8ce78237016cdaa1c54f

## How is that should be
Ideally it should work with the newest django-stubs

## System information

- OS:
- `python` version: 3.10
- `django` version: 3.2.12
- `mypy` version: 0.941
- `django-stubs` version: 1.10.0

henribru:
# I have made things!

## Related issues
- Closes #202


henribru:
Struggling a bit to get CI passing ...

setup.py
```diff
@@ -21,7 +21,7 @@ def find_stub_files(name):
 
 dependencies = [
     "mypy>=0.790",
```
sobolevn, line 23: Should we use the same `mypy` version as in `django-stubs`?

henribru: Sure

henribru:
The remaining failure seems to be caused by `django-stubs` importing `Literal` from `typing`, which isn't supported in 3.7 :thinking:

sobolevn:
Yeah, it should be `typing_extensions`.

setup.py
```diff
@@ -20,8 +20,8 @@ def find_stub_files(name):
     readme = f.read()
 
 dependencies = [
-    "mypy>=0.790",
-    "django-stubs>=1.7.0",
+    "mypy>=0.930,<0.950",
```
dimbleby, line 23: is this upper bound on mypy necessary?  (and does it relate to the purpose of this MR?)

This project releases rather infrequently.  So as a user I'd prefer that it did not pin me to a mypy version that will soon become stale.  

If and when mypy release a breaking change, I can always pin it myself - but the recent evidence (per the current requirement without an upper bound) is that that's probably not needed.

sobolevn: Mypy can break its plugin API in new releases. We do the same in `django-stubs`

dimbleby: Hmm, well if I'd known that I'd have asked you not to do it there too!

Perhaps you know something that I don't about mypy versioning, it seems mysterious to me.  Is there any reason to suspect that 0.950 would be breaking but 0.949 would not?

henribru: > is this upper bound on mypy necessary? (and does it relate to the purpose of this MR?

I just set it to the same one as `django-stubs` as suggested in https://github.com/typeddjango/djangorestframework-stubs/pull/203#discussion_r835755300.

henribru:
> The remaining failure seems to be caused by `django-stubs` importing `Literal` from `typing`, which isn't supported in 3.7 🤔

I've opened a PR against `django-stubs` to fix this

henribru:
The PR against `django-stubs` has been merged, so I guess we're mostly waiting for a new release there

henribru:
Passing now, after fixing another unrelated failure :sweat_smile:

.pre-commit-config.yaml
```diff
@@ -23,6 +23,8 @@ repos:
     rev: 21.12b0
```
sobolevn, line 23: Can you please update `black` to the latest version please? It would also fix this problem.

henribru: Sure, I was just worried it would reformat a whole bunch of stubs. I'll do it later today

henribru: Done. Only resulted in two very minor whitespace changes.

sobolevn (approved):
Great work! Thank you!

PR #205, opened 2022-03-31
Fix "HttpResponse" has no attribute "data" on TestCases

### LINKED ISSUE #204: "HttpResponse" has no attribute "data"
# Bug report

## What's wrong

When running `mypy` with the `--strict` flag on a test it will raise `error: "HttpResponse" has no attribute "data"`

Test example:
```
# api/tests.py (lines 9-12)
class ExampleTest(APITestCase):
    def test_failure(self) -> None:
        response = self.client.get('/hello/', format="json")
        self.assertEqual(response.data["message"], "Hello, world!")
```

## How is that should be

`mypy` should not complain about `response` not having the `data` attribute.

## How to reproduce

Repo to reproduce the issue:  
https://github.com/XF-FW/issue-drf-stubs-response
Github action that fails when running `mypy` due to the issue in question:  
https://github.com/XF-FW/issue-drf-stubs-response/runs/5759647545?check_suite_focus=true

## Why

- Is django-stubs not in the latest version?
  - #202 

## Have you tried:

- Upgrading django-stubs and not using the drf-stubs mypy plugin?
   - Yes. It does not work.
- Using the `master` branch, instead of the lastest release?
   - Yes. It does not work. Although I expected it to work, since something similar is mentioned in the CHANGELOG. For reference: https://github.com/typeddjango/djangorestframework-stubs/blob/master/CHANGELOG.md?plain=1#L10
- Applying the same changes that #105 did?
   - Yes. **It does work!**

## The end.
Thanks for putting up with me. Let me know if I can help in any way.

XF-FW:
# I have made things!

## Related issues

Fixes #204

### DIFF
--- a/rest_framework-stubs/test.pyi
+++ b/rest_framework-stubs/test.pyi
@@ -81,14 +81,18 @@ class APIClient(APIRequestFactory, DjangoClient):
 
 class APITransactionTestCase(testcases.TransactionTestCase):
     client_class: Type[APIClient] = ...
+    client: APIClient
 
 class APITestCase(testcases.TestCase):
     client_class: Type[APIClient] = ...
+    client: APIClient
 
 class APISimpleTestCase(testcases.SimpleTestCase):
     client_class: Type[APIClient] = ...
+    client: APIClient
 
 class APILiveServerTestCase(testcases.LiveServerTestCase):
     client_class: Type[APIClient] = ...
+    client: APIClient
 
 class URLPatternsTestCase(testcases.SimpleTestCase): ...
--- a/tests/typecheck/test_test.yml
+++ b/tests/typecheck/test_test.yml
@@ -1,9 +1,28 @@
-- case: test_api_test_case
+- case: test_testcases_client_api
+  parametrized:
+  - test_class: APITransactionTestCase
+  - test_class: APITestCase
+  - test_class: APISimpleTestCase
+  - test_class: APILiveServerTestCase
   main: |
     from rest_framework import test, status
 
-    class MyTest(test.APITestCase):
+    class MyTest(test.{{ test_class }}):
         def test_example(self) -> None:
+            reveal_type(self.client) # N: Revealed type is "rest_framework.test.APIClient"
             response = self.client.get('/', format="json")
             self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
             self.assertEqual(response.data, {"detail": {"not_found"}})
+
+- case: test_testcases_client_api_urlpatterns
+  main: |
+    from typing import Union, List
+    from django.urls import URLPattern, URLResolver
+    from rest_framework import test, status
+
+    class MyTest(test.URLPatternsTestCase):
+        urlpatterns : List[Union[URLPattern, URLResolver]] = []
+        def test_example(self) -> None:
+            reveal_type(self.client) # N: Revealed type is "django.test.client.Client"
+            response = self.client.get('/', format="json")
+            self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
--- a/tests/typecheck/test_test.yml
+++ b/tests/typecheck/test_test.yml
@@ -1,0 +1,9 @@
+- case: test_api_test_case
+  main: |
+    from rest_framework import test, status
+
+    class MyTest(test.APITestCase):
+        def test_example(self) -> None:
+            response = self.client.get('/', format="json")
+            self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
+            self.assertEqual(response.data, {"detail": {"not_found"}})


XF-FW:
Ok, this is me trying to ensure myself, that this fix is the correct one and to help anyone reading to interpret this as well:

The root of the issue is this error:
`"_MonkeyPatchedWSGIResponse" has no attribute "data"`

I think it's worth mentioning that the error has changed slightly, due to changes made in https://github.com/typeddjango/django-stubs/pull/925.

Anyway, we can see that the revealed type is wrong, by this partial example:
```
response = self.client.get(self.url, format="json")
reveal_type(self.client)  
# Revealed type is "django.test.client.Client"
# Expected type would be "rest_framework.test.APIClient"
```

Ok, let me put out some links here, so we can quickly check whatever I'm writing:

Here we have the definition for the different testcases classes that DRF provides:
https://github.com/encode/django-rest-framework/blob/e5fb9af0eaffde683fa0af3987085f86cf0d2640/rest_framework/test.py#L345  
And the equivalent stubs:
https://github.com/typeddjango/djangorestframework-stubs/blob/61b10355d991be4af2ba8e2457d3ccc33bbae0c7/rest_framework-stubs/test.pyi#L82

The stubs for the parent testcase class that Django provides:
https://github.com/typeddjango/django-stubs/blob/42d8e18bf8d35482cdb4935866bd35304a74fc9d/django-stubs/test/testcases.pyi#L65

And stubs for Django's `Client` class:
https://github.com/typeddjango/django-stubs/blob/42d8e18bf8d35482cdb4935866bd35304a74fc9d/django-stubs/test/client.pyi#L150

Finally, for completeness, for the curious ones, this is where `client` is assigned to an instance of `client_class`
https://github.com/django/django/blob/8c0886b068ba4e224dd78104b93c9638b860b398/django/test/testcases.py#L432

Oof. I promise that's everything.  
Ok, the problem here, seems to be that `django-stubs` declares the following:
```
class SimpleTestCase(unittest.TestCase):
    client_class: Type[Client] = ...
    client: Client
```

Client is typed as:
```
class Client(ClientMixin, _RequestFactory[_MonkeyPatchedWSGIResponse]):
    ...
    def request(self, **request: Any) -> _MonkeyPatchedWSGIResponse: ...
```

The initial error, now makes sense. It's using Django `Client` type for the `client` instance variable, instead of `APIClient`. Let's just check how are the stubs of `APIClient` defined:
```

...

class APIClient(APIRequestFactory, DjangoClient):
    ...
    def request(self, **kwargs: Any) -> Response: ...  # type: ignore[override]
```

Although it's pretty trivial knowledge, but we have our hands dirty, so, 
while we're at it, we can confirm that rest_framework.response.Response has indeed a `data` attribute:
```
class Response(SimpleTemplateResponse):
    data: Any = ...
    ...
```

Anyway, back to reality. The problem is that we're using the parent class for the `client` type. This doesn't allow us to use the methods only defined in the child class. We need to narrow the type. I think that's what I did with this PR, by simply declaring `client` as having the narrower type.

Let's wrap this up, because this has like 100x times the size of the PR.

PS: Wrote an accompanying test.

rest_framework-stubs/test.pyi
```diff
@@ -81,14 +81,18 @@ class APIClient(APIRequestFactory, DjangoClient):
...
 
 class APISimpleTestCase(testcases.SimpleTestCase):
     client_class: Type[APIClient] = ...
+    client: APIClient
 
 class APILiveServerTestCase(testcases.LiveServerTestCase):
     client_class: Type[APIClient] = ...
+    client: APIClient
 
 class URLPatternsTestCase(testcases.SimpleTestCase): ...
```
sobolevn, line 98: What about this case?

tests/typecheck/test_test.yml
```diff
@@ -0,0 +1,9 @@
+- case: test_api_test_case
+  main: |
+    from rest_framework import test, status
+
+    class MyTest(test.APITestCase):
```
sobolevn, line 5: Can you please check other types as well? You can use https://github.com/typeddjango/pytest-mypy-plugins#2-parametrized

sobolevn (commented):
Great explanation ;)

rest_framework-stubs/test.pyi
XF-FW: `URLPatternsTestCase` does not define a custom `client` as far as I can tell:
https://github.com/encode/django-rest-framework/blob/e5fb9af0eaffde683fa0af3987085f86cf0d2640/rest_framework/test.py#L368

tests/typecheck/test_test.yml
XF-FW: Done!

XF-FW: Thanks for the link, btw. It was pretty helpful.

sobolevn (approved):
Thank you!

PR #208, opened 2022-04-13
Fix the APIException detail attribute type

canassa:
DRF does not use raw strings but the ErrorDetail which is a subclass of str

See: https://github.com/encode/django-rest-framework/blob/master/rest_framework/exceptions.py#L18


intgr:
Sorry for not looking at this earlier.

This has been fixed as part of a larger change:
* #387

https://github.com/typeddjango/djangorestframework-stubs/pull/387/files#diff-1c7271d7b37e75f0bc7c5a29463ca7c5d6649959f12085c3c3d3d15c309e1440L18-R14

PR #215, opened 2022-04-29
Preserve generic in extended generic views and viewsets

henribru:
# I have made things!

Without this, you won't be able to specify a type for view sets and things like RetrieveAPIView and the generic argument gets inferred as `Any`.


sobolevn (approved):
Thank you!

PR #216, opened 2022-05-02
Update exceptions.pyi & add test

XF-FW:
Set the `default_detail` as `_Detail`, defined above as `_Detail = Union[str, List[Any], Dict[str, Any]]`. 
The `detail` argument already has that type, so I think this change is fairly safe.

This change was made, to allow something like this without `mypy` complaining.
```
class MyException(exceptions.APIException):
    status_code = 200
    default_detail = {"ok": "everything"} # not a string
    default_code = "ok"
```

Added a couple of tests.


sobolevn (approved):
Thanks!

PR #218, opened 2022-05-05
More strictly hint URL pattern members

intgr:
# I have made things!

In many cases `URLPattern` might be sufficient -- as long as the pattern doesn't use `include()`. But determining where include patterns can or cannot occur is difficult, so using `Union[URLPattern, URLResolver]` everywhere.

Using `Sequence[Union...]` instead of `List[Union...]`, because the former is covariant. Users may be passing an auto-inferred list of patterns, which may be `List[URLPattern]`

## Related issues

* Companion PR to `django-stubs`: https://github.com/typeddjango/django-stubs/pull/949


intgr:
Added `_AnyURL = Union[URLPattern, URLResolver]` alias here, just like django-stubs PR.

intgr:
Also changed a few `Sequence[_AnyURL] -> Iterable[_AnyURL]` based on reviewing djangorestframework code.

sobolevn (approved):
Thank you! 🎉

Issue #145, opened 2021-05-13
Can 'coreapi' dependency be dropped?


intgr:
## What's wrong

Hi! In my project, `djangorestframework-stubs`'s dependency on `coreapi` pulls in several unneeded dependencies, which could otherwise be removed.

```
djangorestframework-stubs 1.4.0 PEP-484 stubs for django-rest-framework
├── coreapi >=2.0.0
│   ├── coreschema * 
│   │   └── jinja2 * 
│   │       └── markupsafe >=0.23 
│   ├── itypes * 
│   ├── requests * 
│   │   ├── certifi >=2017.4.17 
│   │   ├── chardet >=3.0.2,<5 
│   │   ├── idna >=2.5,<3 
│   │   └── urllib3 >=1.21.1,<1.27 
│   └── uritemplate * 
│   ...
```

I don't entirely understand the relationship bertween DRF and `coreapi`, since I haven't used that functionality. But I notice that DRF upstream has been deprecating the dependency, and the latest 3.12 version no longer requires it.

https://www.django-rest-framework.org/community/3.10-announcement/

> This release begins the deprecation process for the CoreAPI based schema generation, and introduces OpenAPI schema generation in its place.

----

I can put in some development work myself, but I don't know where to start yet. So first I wanted to see if the maintainers already have an idea about what needs to be done?

sobolevn:
The thing here is that we don't have any versioning policy yet. In `django-stubs` we at least have this: https://github.com/typeddjango/django-stubs/#version-compatibility

I think, that `coreapi` can be dropped after `drf` drops it.

PR #220, opened 2022-05-06
Fix broken imports and references

intgr:
# I have made things!

* Fixed broken imports and references in `coreapi`.
* Changed mypy config (plugins.ini) to set `ignore_missing_imports = True` per module instead of globally.
* Added missing `__init__.pyi` to `management/` and `management/commands/` folders.
* Cleaned up an unused `MutableMapping` import.

Ugh, did this ever work? :)

## Related issues

* #145


rest_framework-stubs/schemas/coreapi.pyi
```diff
@@ -53,11 +53,11 @@ class AutoSchema(ViewInspector):
     def get_filter_fields(self, path: str, method: str) -> List[Dict[str, Any]]: ...
     def get_manual_fields(self, path: str, method: str) -> List[coreapi.Field]: ...
     @staticmethod
-    def update_fields(fields: List[coreapi.Fields], update_with: List[coreapi.Fields]) -> List[coreapi.Fields]: ...
+    def update_fields(fields: List[coreapi.Field], update_with: List[coreapi.Field]) -> List[coreapi.Field]: ...
```
sobolevn, line 56: Maybe this line is the cause of the problem? https://github.com/typeddjango/djangorestframework-stubs/blob/master/tests/plugins.ini#L2

tests/plugins.ini
```diff
@@ -1,6 +1,20 @@
...
     mypy_django_plugin.main,
     mypy_drf_plugin.main
+
+[mypy-coreapi]
+ignore_missing_imports = True
+
+[mypy-coreschema]
+ignore_missing_imports = True
+
+[mypy-pygments]
```
sobolevn, line 13: This can be covered by https://pypi.org/project/types-Pygments/ in the future.

tests/plugins.ini
```diff
@@ -1,6 +1,20 @@
...
+ignore_missing_imports = True
+
+[mypy-psycopg2.extras]
```
sobolevn, line 16: Possibly https://pypi.org/project/types-psycopg2/ ?

sobolevn (approved):
Thanks! Glad we did fix this. It is strage that this bug was hidden for such a long time.

PR #228, opened 2022-06-07
Add GitHub release action to upload to PyPI & create GitHub release

terencehonles:
Adds a release action like in https://github.com/typeddjango/django-stubs/pull/980


sobolevn (approved):
Thanks!

PR #237, opened 2022-06-22
Use ImmutableQueryDict for request params

intgr:
# I have made things!

See https://github.com/typeddjango/django-stubs/pull/925#discussion_r851922533 for some background. django-stubs itself uses `_ImmutableQueryDict` for query params, but djangorestframework-stubs was not updated with this.

`_ImmutableQueryDict` was added in django-stubs 0.11.0, which is already a requirement in djangorestframework-stubs.

## Related issues

Closes https://github.com/typeddjango/django-stubs/issues/1010

### DIFF
--- a/tests/typecheck/test_request.yml
+++ b/tests/typecheck/test_request.yml
@@ -1,8 +1,7 @@
 -   case: request_querydict
     main: |
-        from typing_extensions import assert_type
         from rest_framework.request import Request
 
         def some_view(request: Request) -> None:
-            assert_type(request.query_params['field'], str)
-            assert_type(request.POST['field'], str)
+            reveal_type(request.query_params['field'])  # N: Revealed type is "builtins.str"
+            reveal_type(request.POST['field'])          # N: Revealed type is "builtins.str"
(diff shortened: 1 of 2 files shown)


intgr:
Cc: @sterliakov

tests/typecheck/test_request.yml
```diff
@@ -0,0 +1,8 @@
+-   case: request_querydict
+    main: |
+        from typing_extensions import assert_type
+        from rest_framework.request import Request
+
+        def some_view(request: Request) -> None:
+            assert_type(request.query_params['field'], str)
+            assert_type(request.POST['field'], str)
```
sobolevn, line 8: Please, let's use `reveal_type` and `# N: ` comment. This is the style we use in these projects. Type tests are confusing, let's not make them even more confusing ;)

intgr: I think `assert_type` is superior in many respects to the current style. But fair enough, not the topic of this PR.

sobolevn:
Thanks, this looks good enough to me for now!

sobolevn:
Sadly, it didn't make it into 1.7.0
I will release 1.7.1 right now.

PR #263, opened 2022-10-04
Add missing PageNumberPagination.get_page_number() method

intgr:
Added missing method


intgr:
The CI seems broken? Same output from PR #262 https://github.com/typeddjango/djangorestframework-stubs/actions/runs/3180319143

sobolevn:
Yes, looks like so.
If you have time to fix it, this would be very much appricated.
If not - no problem, I will get to it sometime soon.

sobolevn (approved):

sobolevn:
Thank you!

PR #267, opened 2022-10-18
Fix type of data parameter in get test requests

nils-van-zuijlen:
Lists of two-tuples of strings are allowed in the data attribute for get requests

Example of a false-positive needing this fix:
```python
        factory = ApiRequestFactory()
        request = factory.get(users_list_url, data=[("manager", manager1.username), ("manager", manager2.username)])
```

### DIFF
--- a/rest_framework-stubs/test.pyi
+++ b/rest_framework-stubs/test.pyi
@@ -1,4 +1,6 @@
+from collections.abc import Iterable, Mapping
 from typing import Any
+from typing_extensions import TypeAlias
 
 import coreapi
 import requests
@@ -13,6 +15,12 @@ from django.test.client import RequestFactory as DjangoRequestFactory
 from rest_framework.authtoken.models import Token
 from rest_framework.request import Request
 from rest_framework.response import _MonkeyPatchedResponse
+
+_GetDataType: TypeAlias = (
+    Mapping[str, str | bytes | int | Iterable[str | bytes | int]]
+    | Iterable[tuple[str, str | bytes | int | Iterable[str | bytes | int]]]
+    | None
+)
 
 def force_authenticate(
     request: HttpRequest, user: AnonymousUser | AbstractBaseUser | None = ..., token: Token | None = ...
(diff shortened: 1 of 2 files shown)


nils-van-zuijlen:
The import paths seem to be wrong in the original code, hence why the tests are broken.

intgr:
CI fixed again, sorry! I rebased your branch.

sobolevn (approved):
LGTM!

rest_framework-stubs/test.pyi
```diff
@@ -14,6 +15,8 @@ from rest_framework.authtoken.models import Token
 from rest_framework.request import Request
 from rest_framework.response import _MonkeyPatchedResponse
 
+_GetDataType: TypeAlias = dict[str, Any] | str | list[tuple[str, str]] | None
```
intgr, line 18: Hmm. `str` was an allowed input before, but looking at the code `django.utils.http.urlencode`, passing `str` to `APIRequestFactory.get` seems it will just crash? Unless I'm missing something

And the value is not just `str`, it can be `str | bytes | Iterable[str | bytes]`

And in function inputs, abstract base classes like `Iterable[]` or `Sequence[]` are preferred instead of `list[]`

```suggestion
_GetDataType: TypeAlias = Mapping[str, str | bytes | Iterable[str | bytes]] | Iterable[tuple[str, str | bytes | Iterable[str | bytes]]] | None
```

intgr (changes requested):
Thanks for the PR

intgr:
A similar change would be helpful in `django-stubs` as well for the `RequestFactory` class and `django.utils.http.urlencode` function.

intgr:
The lint and test are failing because an `import` is missing for `Mapping`, `Iterable`

But typecheck fails due to changes in upstream DRF tests; I'll fix that in a bit.

intgr:
From typecheck output

> drf_source/tests/test_pagination.py:546: error: Argument 2 to "get" of "APIRequestFactory" has incompatible type "Dict[str, int]"; expected "Union[Mapping[str, Union[str, bytes, Iterable[Union[str, bytes]]]], Iterable[Tuple[str, Union[str, bytes, Iterable[Union[str, bytes]]]]], None]"

Maybe we should accept `int` as well? It's weird to pass URL params as integer, but on the other hand, this does work and is covered by unit tests.

If there are still remaining typecheck errors after that, we should just suppress them via `IGNORED_ERRORS`.

intgr (approved):
Sorry, I've neglected this project a bit while I focused on django-stubs. Rest assured, I haven't forgotten about it. :)

intgr:
I rebased your branch, there are some false positives in CI typecheck which shoud probably be added to ignores list.

intgr (approved):
I pushed the relevant typecheck ignores and this can now be merged. Thanks!

PR #273, opened 2022-11-03
Fix errors in CI

intgr:
Thanks to @andersk for related fix in `django-stubs`: https://github.com/typeddjango/django-stubs/pull/1123


intgr:
@sobolevn Looks like I did it. We're not yet compatible with `django-stubs 1.13.0`, but taking it one step at a time.

sobolevn (approved):
Thank you!

PR #275, opened 2022-11-06
Update mypy to 0.982

emicba:
# I have made things!

## Related issues

Closes #274

PR #279, opened 2022-11-08
Update django-stubs, mypy, pytest-mypy-plugins & fix tests

### LINKED ISSUE #274: Package cannot be installed along with `django-stubs==1.13.0`
# Bug report

Due to incompatible version constraint on `mypy` dependency, `djangorestfamework-stubs` cannot be installed with `django-stubs==1.13.0`.

## What's wrong

Example error from `poetry`'s resolver:
```
  SolverProblemError

  Because djangorestframework-stubs (1.7.0) depends on mypy (>=0.950,<0.970)
   and django-stubs (1.13.0) depends on mypy (>=0.980,<0.990), djangorestframework-stubs (1.7.0) is incompatible with django-stubs (1.13.0).
  So, because dcmanager depends on both django-stubs (1.13.0) and djangorestframework-stubs (1.7.0), version solving failed.
```

## How is that should be

I guess releasing a patch version of `djangorestframework-stubs` with the constraint on `mypy` matching the one of `django-stubs==1.13.0`, but I don't know what your release policy is.

## System information

- OS: doesn't matter, should be reproducible under any OS
- `python` version: 3.10
- `django` version: 4.1.3
- `mypy` version: 0.961
- `django-stubs` version: 1.13.0

intgr:
All these updates have to be done together, otherwise one dependency is incompatible with another in some way that fails in CI.

Fixes #275, fixes #274

### DIFF
--- a/scripts/drf_tests_settings.py
+++ b/scripts/drf_tests_settings.py
@@ -1,7 +1,7 @@
 SECRET_KEY = "1"
 SITE_ID = 1
 
-INSTALLED_APPS = [
+INSTALLED_APPS = (
     "django.contrib.contenttypes",
     "django.contrib.sites",
     "django.contrib.sessions",
@@ -11,4 +11,4 @@ INSTALLED_APPS = [
     "django.contrib.auth",
     "rest_framework",
     "rest_framework.authtoken",
-]
+)
--- a/scripts/typecheck_tests.py
+++ b/scripts/typecheck_tests.py
@@ -76,6 +76,7 @@ IGNORED_ERRORS = {
         'Incompatible types in assignment (expression has type "AsView[GenericView]", variable has type "AsView[Callable[[HttpRequest], Any]]")',  # noqa: E501
         'Argument "patterns" to "SchemaGenerator" has incompatible type "List[object]"',
         'Argument 1 to "field_to_schema" has incompatible type "object"; expected "Field[Any, Any, Any, Any]"',
+        'Argument "help_text" to "CharField" has incompatible type "_StrPromise"',
     ],
     "browsable_api": [
         '(expression has type "List[Dict[str, Dict[str, int]]]", base class "GenericAPIView" defined the type as "Union[QuerySet[_MT?], Manager[_MT?], None]")',  # noqa: E501
--- a/setup.py
+++ b/setup.py
@@ -19,8 +19,8 @@ with open("README.md") as f:
     readme = f.read()
 
 dependencies = [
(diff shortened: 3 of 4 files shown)


scripts/drf_tests_settings.py
```diff
@@ -1,7 +1,7 @@
 SECRET_KEY = "1"
 SITE_ID = 1
 
-INSTALLED_APPS = [
+INSTALLED_APPS = (
     "django.contrib.contenttypes",
     "django.contrib.sites",
```
intgr, line 6: INSTALLED_APPS must be tuple for typecheck_tests, because DRF test suite does:
```python
settings.INSTALLED_APPS += ('guardian',)
```

Better than ignoring IMO.

scripts/typecheck_tests.py
```diff
@@ -76,6 +76,7 @@
         'Incompatible types in assignment (expression has type "AsView[GenericView]", variable has type "AsView[Callable[[HttpRequest], Any]]")',  # noqa: E501
         'Argument "patterns" to "SchemaGenerator" has incompatible type "List[object]"',
         'Argument 1 to "field_to_schema" has incompatible type "object"; expected "Field[Any, Any, Any, Any]"',
+        'Argument "help_text" to "CharField" has incompatible type "_StrPromise"',
```
intgr, line 79: I suppose we can make `help_text` `StrOrPromise` in a follow-up PR.

setup.py
```diff
@@ -20,15 +20,15 @@ def find_stub_files(name):
 
 dependencies = [
     "mypy>=0.950",
-    "django-stubs>=1.11.0",
+    "django-stubs>=1.12.0",
```
intgr, line 23: I'm unsure how much I should bump these lower bound versions. Quite clearly these don't work in our CI, but in user projects they **may** work together?

If we want to be sure I suppose we should require mypy>=0.980 and django-stubs>=1.13.0.

sobolevn: I think we can do whatever is easier for us to maintain :)

intgr: Yeah after some thought, I would bump both of these.

intgr:
@sobolevn I'd like to get this merged, so djangorestframework-stubs is a step closer to being release-able again.

(The other PR for mypy 0.990 can wait)

setup.py
intgr: Bumped these version requirements.

sobolevn (approved):
Great work, thank you!

PR #280, opened 2022-11-08
Update to mypy 0.991 for compatible-mypy & CI

intgr:


mypy.ini
```diff
@@ -2,10 +2,11 @@
 strict_optional = True
 ignore_missing_imports = True
 check_untyped_defs = True
-warn_no_return = False
 show_traceback = True
 allow_redefinition = True
 incremental = True
+show_error_codes = False
+disable_error_code = empty-body
```
intgr, line 9: `warn_no_return` is no longer necessary now that we disable `empty-body`.

rest_framework-stubs/permissions.pyi
```diff
@@ -44,7 +44,11 @@ class NOT(_SupportsHasPermission):
     def __init__(self, op1: _SupportsHasPermission) -> None: ...
 
 class BasePermissionMetaclass(OperationHolderMixin, type): ...  # type: ignore[misc]
-class BasePermission(_SupportsHasPermission, metaclass=BasePermissionMetaclass): ...  # type: ignore[misc]
+
+class BasePermission(metaclass=BasePermissionMetaclass):  # type: ignore[misc]
+    def has_permission(self, request: Request, view: APIView) -> bool: ...
+    def has_object_permission(self, request: Request, view: APIView, obj: Any) -> bool: ...
```
intgr, line 50: If I kept `_SupportsHasPermission` here, it causes "Metaclass conflict" errors.

I don't understand metaclasses well enough to say whether that's a bug in mypy or a genuine issue.

But duplicating these method signatures here fixes it.

intgr:
- [ ] typecheck-tests is not running correctly any more. All ignores show up in "UNUSED IGNORES" now.

scripts/typecheck_tests.py
```diff
@@ -311,7 +316,8 @@ def update(self, op_code, cur_count, max_count=None, message=""):
         dir_util.copy_tree(str(STUBS_DIRECTORY), str(DRF_SOURCE_DIRECTORY / "rest_framework"))
     mypy_config_file = (PROJECT_DIRECTORY / "mypy.ini").absolute()
     mypy_cache_dir = PROJECT_DIRECTORY / ".mypy_cache"
-    tests_root = DRF_SOURCE_DIRECTORY / "tests"
+    # HACK: Work around https://github.com/python/mypy/issues/14042
+    tests_root = Path("drf_source") / "tests"
```
intgr, line 320: To be reverted

sobolevn (approved):
Thanks!

intgr:
Oops @sobolevn I wanted to delay merging this to avoid creating a new conflict for #286. Probably should have left a comment about that. But hopefully this doesn't conflict.

PR #283, opened 2022-11-10
Fixes #230 - Response stub should inherit from TemplateResponse

### LINKED ISSUE #230: error: "Response" has no attribute "json"
# Bug report

## What's wrong

Given the snippet such as this

```python3
from rest_framework.test import APIClient


def test_response(api_client: APIClient) -> None:
    response: Response = api_client.get('https://example.com/example.json')
    assert response.json() == {}
```

and running `mypy .` on it, I receive `error: "Response" has no attribute "json"`
Running the test works fine.

The reason is that `rest_framework-stubs/response.pyi` declares `Response` to inherit from `django.template.response.SimpleTemplateResponse` and this class in turn does not declare the `json` field, it's the `TemplateResponse` that does.

This was already reported in django-stubs here: https://github.com/typeddjango/django-stubs/issues/962
I'm not that person.

## How is that should be

Since the test works mypy shouldn't complain that the method `Response.json()` does not exist.

## System information

- OS: Linux
- `python` version: 3.8.11
- `django` version: 3.0.14
- `mypy` version: 0.950
- `django-stubs` version: 1.11.0

mattwwarren:
# I have made things!

## Related issues

Closes #230 

I'm not absolutely certain the test is in the right place but it does appear to exercise the changed code. Thank you in advance for any feedback!


intgr:
This isn't strictly correct, the class hierarchy still has `SimpleTemplateResponse` as base. During normal view processing, the `Response` class doesn't have a `json()` method either.

I think there's a better solve this, which `django-stubs` implements. It's sort of a hack, but it mirrors the hack that Django does. We would need a new type like `_MonkeyPatchedResponse` like this: https://github.com/typeddjango/django-stubs/blob/master/django-stubs/test/client.pyi#L123-L142

And then make `APIClient` methods return this new type instead of plain `Response`.

mattwwarren:
Thanks for the feedback @intgr 

I'm a little out of my comfort zone but I think I've implemented the `_MonkeyPatchedResponse` in the way you mean.

rest_framework-stubs/response.pyi
```diff
@@ -23,3 +23,6 @@ class Response(SimpleTemplateResponse):
     def render(self) -> Any: ...
     @property
     def status_text(self) -> str: ...
+
+class _MonkeyPatchedResponse(Response):
+    def json(self) -> Any: ...
```
intgr, line 28: I think the other attributes from `_MonkeyPatchedWSGIResponse` are relevant too. Because it seems to re-use the same logic from Django: https://github.com/django/django/blob/9bd174b9a75299dce33e673a559f2b673399b971/django/test/client.py#L818-L824

rest_framework-stubs/response.pyi
```diff
@@ -23,3 +27,12 @@ class Response(SimpleTemplateResponse):
     def render(self) -> Any: ...
     @property
     def status_text(self) -> str: ...
+
+class _MonkeyPatchedResponse(Response):
+    client: DjangoClient
```
intgr, line 32: This can be `APIClient` here

mattwwarren: I wondered about that. What prevents a circular dependency?

intgr: I believe circular dependencies are allowed in `.pyi` stubs files.

mattwwarren: Quite right. TIL

intgr:
✅ This looks good to me.

I'm unable to trigger CI or approve PRs in this repository yet, but hopefully soon. :)

mattwwarren:
> white_check_mark This looks good to me.
> 
> I'm unable to trigger CI or approve PRs in this repository yet, but hopefully soon. :)

Thank you for your review! This should unblock our mypy upgrades so hopefully it can be approved soon :crossed_fingers:

intgr:
Ping @sobolevn, I don't have privileges to approve running CI or to approve PRs in this repository.

sobolevn:
@intgr this is too complex! Sorry, I will make sure that you have needed permissions.

intgr (approved):

PR #284, opened 2022-11-10
CI: Enable testing with Python 3.11

intgr:


intgr:
Python 3.11 test failures should be fixed by https://github.com/typeddjango/django-stubs/pull/1223.

intgr:
Actually, merging this myself as a trivial change.

Issue #238, opened 2022-06-22
Support python3.10


sobolevn:
Right now `typecheck` tests fail with: 

```
drf_source/tests/test_permissions.py:548: error: "object" not callable
drf_source/tests/test_permissions.py:554: error: "object" not callable
drf_source/tests/test_permissions.py:608: error: "object" not callable
drf_source/tests/test_permissions.py:615: error: Cannot assign multiple types to name "composed_perm" without an explicit "Type[...]" annotation
drf_source/tests/test_permissions.py:628: error: "object" not callable
drf_source/tests/test_permissions.py:635: error: Cannot assign multiple types to name "composed_perm" without an explicit "Type[...]" annotation
```
Link: https://github.com/typeddjango/djangorestframework-stubs/runs/7000977052?check_suite_focus=true#step:6:421
Source: https://github.com/encode/django-rest-framework/blob/fa9d516ee24b2447a351af92e139d386b4f0d2b4/tests/test_permissions.py#L548
Initial attempt: https://github.com/typeddjango/djangorestframework-stubs/pull/236

intgr:
Fixed in #292.

PR #292, opened 2022-11-16
CI: Enable testing with Python 3.10

intgr:
Also updated pre-commit to use Python 3.10.

Split out from #284.


intgr:
Merging this myself as trivial.

intgr:
Closes #238.

PR #297, opened 2022-11-18
[Fix] Allow StrPromise to be used in exceptions

XF-FW:
# I have made things!

## Related issues

I didn't create an issue, but it's the mirror of https://github.com/typeddjango/django-stubs/issues/1137, for this repo. There might be more places where the same should be applied, but this resolves my issue.

### DIFF
--- a/rest_framework-stubs/exceptions.pyi
+++ b/rest_framework-stubs/exceptions.pyi
@@ -4,7 +4,7 @@ from typing import Any
 from typing_extensions import TypeAlias
 
 from django.http import HttpRequest, JsonResponse
-from django.utils.functional import _StrPromise
+from django_stubs_ext import StrOrPromise
 from rest_framework.renderers import BaseRenderer
 from rest_framework.request import Request
 
@@ -16,7 +16,7 @@ class ErrorDetail(str):
     code: str | None
     def __new__(cls, string: str, code: str | None = ...): ...
 
-_Detail: TypeAlias = str | _StrPromise | list[Any] | dict[str, Any]
+_Detail: TypeAlias = StrOrPromise | list[Any] | dict[str, Any]
 
 class APIException(Exception):
     status_code: int
--- a/rest_framework-stubs/exceptions.pyi
+++ b/rest_framework-stubs/exceptions.pyi
@@ -4,6 +4,7 @@ from typing import Any
 from typing_extensions import TypeAlias
 
 from django.http import HttpRequest, JsonResponse
+from django.utils.functional import _StrPromise
 from rest_framework.renderers import BaseRenderer
 from rest_framework.request import Request
 
@@ -15,7 +16,7 @@ class ErrorDetail(str):
     code: str | None
     def __new__(cls, string: str, code: str | None = ...): ...
 
-_Detail: TypeAlias = str | list[Any] | dict[str, Any]
+_Detail: TypeAlias = str | _StrPromise | list[Any] | dict[str, Any]
 
 class APIException(Exception):
     status_code: int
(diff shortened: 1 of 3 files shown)


sobolevn (approved):
LGTM, please add new ignores

rest_framework-stubs/exceptions.pyi
```diff
@@ -15,7 +16,7 @@ class ErrorDetail(str):
     code: str | None
     def __new__(cls, string: str, code: str | None = ...): ...
 
-_Detail: TypeAlias = str | list[Any] | dict[str, Any]
+_Detail: TypeAlias = str | _StrPromise | list[Any] | dict[str, Any]
```
intgr, line 19: Don't import private types from django-stubs. Use `StrPromise` from `django_stubs_ext`.

intgr:
The CI isssue will be hopefully fixed in https://github.com/typeddjango/djangorestframework-stubs/pull/303

rest_framework-stubs/exceptions.pyi
intgr: Although I notice we currently don't have an explicit dependency on django-stubs-ext in `setup.py`. IMO that should that be added, thoughts @sobolevn?

intgr:
I rebased this and the CI now passes.

rest_framework-stubs/exceptions.pyi
intgr: Also this could use the `StrOrPromise` type alias:

```suggestion
_Detail: TypeAlias = StrOrPromise | list[Any] | dict[str, Any]
```

sobolevn: `django-stubs-ext` is a dependency of `django-stubs`. I don't think we need to duplicate it 🤔

mschoettle: Sorry to hijack this discussion but I can't find any specific documentation anywhere about this. Just want to confirm that when I have code that uses `gettext` lazy strings I should use either `StrPromise` or `StrOrPromise` (depending on the specific case) from `django-stubs-ext`.

intgr: Sorry for not getting back to you earlier.

`StrOrPromise` accepts both `str` and `StrPromise`.

`StrPromise` only allows `StrPromse`, but not regular `str`.

In nearly all cases `StrOrPromise` is the correct one.

sobolevn (approved):

XF-FW:
I seem to have missed the notifications on this PR.

I'll make the changes that @intgr has suggested, so please hold off merging for a bit.

To clarify, it's replacing `str | _StrPromise` with `StrOrPromise`, is that correct?

intgr:
Yep 👍

rest_framework-stubs/exceptions.pyi
```diff
@@ -4,6 +4,7 @@ from typing import Any
 from typing_extensions import TypeAlias
 
 from django.http import HttpRequest, JsonResponse
+from django.utils.functional import _StrPromise
```
sobolevn, line 7: Oh my, looks like I have confused `_StrPromise` with `_StrOrPormise` as well 🤦 

Naming!

mschoettle:
Thanks! I ran into the same issue when doing the following:

```py
raise serializers.ValidationError(_('Foo'))
```

This PR fixes this issue.

Question regarding the use of `django-stubs-ext`: Does this require `django-stubs-ext` to be added to the dependencies in `setup.py`?

intgr:
> Question regarding the use of django-stubs-ext: Does this require django-stubs-ext to be added to the dependencies in setup.py? (based on this review in another MR this is necessary: https://github.com/typeddjango/djangorestframework-stubs/pull/301#discussion_r1038762553)

I don't have strong feelings about that. In some other PR comments, sobolevn felt that it's not necessary. django-stubs aready depends on django-stubs-ext and as long as that remains, this will work. Let's leave it be.

PR #303, opened 2022-12-02
Fix CI: add typecheck suppression & upgrade flake8-pyi

intgr:
* Added suppression for a new test in upstream
* Upgraded flake8-pyi to be compatible with flake8 v6:

```
    Collecting flake8-pyi==22.10.0
      Downloading flake8_pyi-22.10.0-py37-none-any.whl (24 kB)
    
    The conflict is caused by:
        The user requested flake8 6.0.0 (from /home/runner/.cache/pre-commit/repovn0unijf)
        flake8-pyi 22.10.0 depends on flake8<6.0.0 and >=3.2.1
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,11 +24,11 @@ repos:
     hooks:
       - id: black
   - repo: https://github.com/PyCQA/flake8
-    rev: 5.0.4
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Thanks!

PR #304, opened 2022-12-02
Fix encoder and decoder params on JSONField

zetahernandez:
# I have made things!

## Related issues

params encoder and decoder on JSONField are class types.
Similar example on django_stubs 
https://github.com/typeddjango/django-stubs/blob/b81b1bf3868eaab9fecebdd56ed817ffcb07fad8/django-stubs/db/models/fields/json.pyi#L14-L15


intgr (approved):
Thanks! Looks good. Will merge after the CI fix PR is also merged.

intgr:
Thanks! I rebased your branch just to verify CI and merged.

PR #315, opened 2023-01-08
Use `__new__` instead of `__init__` in serializers

### LINKED ISSUE #260: Incompatible type with `many=True`
# Bug report

## What's wrong

```python3
class MySerializer(serializers.Serializer[int]): ...

MySerializer([1,2,3], many=True)
```

```
error: Argument 1 to "MySerializer" has incompatible type "List[int]"; expected "Optional[int]"  [arg-type]
```

## How is that should be

```
Success: no issues found in 1 source file
```

## System information

- OS: `Linux f16759b1b324 5.10.104-linuxkit #1 SMP PREEMPT Thu Mar 17 17:05:54 UTC 2022 aarch64 GNU/Linux`
- `python` version: 3.10.6
- `django` version: 4.0.7
- `mypy` version: 0.961
- `django-stubs` version: 1.12.0

Also, it isn't asked for by the template, but:

- `djangorestframework` version: 3.13.1
- `djangorestframework-stubs` version: 1.7.0

Viicos:
Attempt to fix #260.


rest_framework-stubs/serializers.pyi
```diff
@@ -81,13 +81,57 @@ class BaseSerializer(Generic[_IN], Field[Any, Any, Any, _IN]):
...
+        default: Any = ...,
+        initial: Any = ...,
+        source: str = ...,
+        label: str = ...,
+        help_text: str = ...,
+        style: dict[str, Any] = ...,
+        error_messages: dict[str, str] = ...,
+        validators: Sequence[Validator[Any]] | None = ...,
+        allow_null: bool = ...,
+    ) -> ListSerializer[_IN]: ...
```
Viicos, line 107: I don't think there's a way of inferring the correct subclass of `ListSerializer`, as it is [defined](https://www.django-rest-framework.org/api-guide/serializers/#customizing-listserializer-behavior) in the `Meta` class

intgr:
This has been in the back of my mind also. It would be great if this works.

AFAIK `def __new__(...) -> ListSerializer` is not allowed: mypy requires `__new__` return type to be a subclass of the class being defined. There's a discussion about this somewhere in mypy issues or PRs, I couldn't find it right now.

But I think it's possible to convince mypy developers to change this requirement.

Please add a test case demonstrating whether this works or not.

intgr:
The CI errors will be fixed in #316.

Viicos:
> AFAIK `def __new__(...) -> ListSerializer` is not allowed: mypy requires `__new__` return type to be a subclass of the class being defined. There's a discussion about this somewhere in mypy issues or PRs, I couldn't find it right now.

Well `ListSerializer` is a subclass of `BaseSerializer`, so this should be alright I guess. Will add tests

Viicos:
Ok this seems to work fine when using `BaseSerializer`:

```python
IntBaseSerializer = serializers.BaseSerializer[int]
test_serializer = IntBaseSerializer([1, 2], many=True)
reveal_type(test_serializer) # N: Revealed type is "rest_framework.serializers.ListSerializer[builtins.int]"
```

But when defining a custom serializer:

```python
class TestSerializer(serializers.Serializer[int]):
    pass

test_serializer = TestSerializer(instance=[1, 2], many=True)
reveal_type(test_serializer) # N: Revealed type is "main.TestSerializer"
```
We get the wrong type, possibly because mypy isn't using the `__new__` overload defined in `BaseSerializer`. But I suspect this might be because `ListSerializer` isn't a subclass of `TestSerializer`, so mypy isn't giving the correct type.

I did some tests locally, and it seems that this issue is specific to mypy:

```python
from __future__ import annotations

class A:
    def __new__(cls) -> B:
        return B()

class B(A):
    pass

class C(A):
    pass

a = C()
reveal_type(a) # Revealed type is "C", but got B with pyright (Pylance)
```

I'll open a ticket `@mypy`, and I think this could be an opportunity to discuss what you said @intgr about the requirement that the return type must be a subclass of the class being defined

intgr:
I found there's already an open mypy issue about this: https://github.com/python/mypy/issues/8330

PR #311, opened 2022-12-21
Fix CI typecheck (add new ignore)

intgr:
Refs #267.

PR #320, opened 2023-01-19
Allow subclasses of APIView to override get_permissions with return type implementing _SupportsHasPermission

### LINKED ISSUE #319: When overriding APIView.get_permissions, mypy complains about return type incompatebility
In this example:

```python
from rest_framework.viewsets import GenericViewSet
from rest_framework.permissions import BasePermission


class MyView(GenericViewSet):
    def get_permissions(self) -> list[BasePermission]:
        ...
```
Mypy generates the following error:
```
example.py:6: error: Return type "List[BasePermission]" of "get_permissions" incompatible with return type "List[_SupportsHasPermission]" in supertype "APIView"  [override]
```
Even though `BasePermission` implements `_SupportsHasPermission`

## System information

- OS:
- `python` version: 3.10
- `django` version: 4.1
- `mypy` version: 0.991
- `django-stubs` version: 1.13

guldfisk:
Closes #319
Issue is that `list` is invariant, which makes the protocol not work.
Changing it so a sequence of course breaks stuff that relies on the result being mutable, but I am assuming that that should not be supported anyways.


intgr (approved):
Thanks, this makes sense.

intgr:
The `SimpleRouter` errors are unrelated to this branch, they should be fixed as a separate PR (similar to #311). I can look into it later if you don't get to it first.

PR #322, opened 2023-01-19
Add missing allow_empty arg to ManyRelatedField

### LINKED ISSUE #321: allow_empty arg missing from ManyRelatedField
When checking
```python
from rest_framework.serializers import ManyRelatedField
from rest_framework.serializers import Serializer


    profile_uploads = ManyRelatedField(allow_empty=True)
```
with mypy, it throws the error:
```
example.py:6: error: Unexpected keyword argument "allow_empty" for "ManyRelatedField"  [call-arg]
```
This arg should be allowed though: https://github.com/encode/django-rest-framework/blob/0618fa88e1a8c2cf8a2aab29ef6de66b49e5f7ed/rest_framework/relations.py#L493

## System information

- OS:
- `python` version: 3.10
- `django` version: 4.1
- `mypy` version: 0.991
- `django-stubs` version: 1.13

guldfisk:
Closes #321

### DIFF
--- a/rest_framework-stubs/relations.pyi
+++ b/rest_framework-stubs/relations.pyi
@@ -177,6 +177,7 @@ class ManyRelatedField(Field[Sequence[Any], Sequence[Any], list[Any], Any]):
         error_messages: dict[str, str] | None = ...,
         validators: Sequence[Validator[Sequence[Any]]] | None = ...,
         allow_null: bool = ...,
+        allow_empty: bool = ...,
         child_relation: RelatedField = ...,
     ): ...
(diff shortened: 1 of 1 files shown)


intgr (approved):
Thanks! For changes like these, ther's no need to bother creating an issue, just a PR is fine.

intgr:
Thanks! CI passed after rebasing.

Looking at DRF code, I noticed that `allow_empty` can only be passed as a kwarg.

If you're up for it, would be nice to add `*,` separating positional and keyword-only arguments: https://peps.python.org/pep-3102/ Requires some reverse engineering in DRF code though.

PR #323, opened 2023-01-19
Fix CI typecheck (add new ignores)

intgr:
Suppressing new typecheck errors that were exposed in #320.

Now ignoring messages like 'note: "SimpleRouter" defined here' and "note: 'IntegerSerializer" defined here' globally because these aren't actually errors, just context.

https://github.com/typeddjango/djangorestframework-stubs/actions/runs/3957376234/jobs/6777713016


intgr:
Merging this myself as trivial change.

PR #330, opened 2023-01-26
Add return type to APIClient.get method

### LINKED ISSUE #327: Missing return type in APIClient get method
# Bug report
The test `APIClient`'s `get` method is missing the return type. I tried to get feedback from the support channels in the README, but it doesn't seem like they are very active.

## What's wrong
The definition of the `APIClient` class in the `rest_framework-stubs/test.pyi` file is missing it's return type (assumed to be `_MonkeyPatchedResponse` like all other methods).

## How is that should be
```py
# rest_framework-stubs/test.pyi
# ...
class APIClient(APIRequestFactory, DjangoClient):
    # ...
    def get(self, path: str, data: dict[str, Any] | str | None = ..., follow: bool = ..., **extra: Any) -> _MonkeyPatchedResponse: ...  # type: ignore[override]
    # ...
```

## System information

- OS: -
- `python` version: 3.11
- `django` version: 4.1.5
- `mypy` version: 0.991
- `django-stubs` version: 1.13.1
- `djangorestframework-stubs` version: 1.8.0

piotrszyma:
# I have made things!

Added return type to `rest_framework.test.APIClient`.

Typing for this class was added in https://github.com/typeddjango/djangorestframework-stubs/commit/cf9d492ec01f99e9a3fa89558aa1fbf1a749bc22 and return type was already missing.

Did not find a reason why it should not be there?

## Related issues

Closes #327

### DIFF
--- a/tests/typecheck/test_test.yml
+++ b/tests/typecheck/test_test.yml
@@ -33,9 +33,16 @@
     from django.urls import URLPattern, URLResolver
     from rest_framework import test, status
 
-    class MyTest(test.APITestCase):
+    client = test.APIClient()
+    response = client.get('/', format="json")
+    reveal_type(response) # N: Revealed type is "rest_framework.response._MonkeyPatchedResponse"
 
-        def test_example(self) -> None:
-            response = self.client.get('/', format="json")
-            reveal_type(response) # N: Revealed type is "rest_framework.response._MonkeyPatchedResponse"
-            self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
+- case: test_api_client_factory_get
+  main: |
+    from typing import Union, List
+    from django.urls import URLPattern, URLResolver
+    from rest_framework import test, status
+
+    factory = test.APIRequestFactory()
+    request = factory.get("/")
+    reveal_type(request) # N: Revealed type is "rest_framework.request.Request"
--- a/tests/typecheck/test_test.yml
+++ b/tests/typecheck/test_test.yml
@@ -26,19 +26,3 @@
             reveal_type(self.client) # N: Revealed type is "django.test.client.Client"
             response = self.client.get('/', format="json")
             self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
-
-- case: test_api_client_get
-  main: |
-    from rest_framework import test
-
-    client = test.APIClient()
-    response = client.get('/', format="json")
-    reveal_type(response) # N: Revealed type is "rest_framework.response._MonkeyPatchedResponse"
-
-- case: test_api_client_factory_get
-  main: |
-    from rest_framework import test
-
-    factory = test.APIRequestFactory()
-    request = factory.get("/")
-    reveal_type(request) # N: Revealed type is "rest_framework.request.Request"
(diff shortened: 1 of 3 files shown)


PedroPerpetua:
Possibly unrelated, but the APIFactory above is also missing it's return type:
```py
# rest_framework-stubs/test.py
class APIRequestFactory(DjangoRequestFactory):
    # ...
    def get(self, path: str, data: dict[str, Any] | str | None = ..., follow: bool = ..., **extra: Any): ...  # type: ignore[override]
    # ...
```
May be worth adding; looking at the surrounding code, it's likely it should be `Request`.

piotrszyma:
Good catch @PedroPerpetua  - added typing for `APIRequestFactory`. Maybe there is some reason why for the return type was missing for the `get` method? 🤔

PedroPerpetua:
> Good catch @PedroPerpetua - added typing for `APIRequestFactory`. Maybe there is some reason why the return type was missing for the `get` method? 🤔

That was my thought too and why I didn't open a PR - was hoping someone at DRF-stubs could answer about it.

piotrszyma:
It looks like some DRF test cases started to fail due to using fake implementations not inheriting original classes. I've added these `mypy` errors to ignore.

```
drf_source/tests/test_reverse.py:43: error: Incompatible types in assignment (expression has type "MockVersioningScheme", variable has type "Optional[BaseVersioning]")
drf_source/tests/test_reverse.py:50: error: Incompatible types in assignment (expression has type "MockVersioningScheme", variable has type "Optional[BaseVersioning]")
drf_source/tests/test_versioning.py:136: error: Incompatible types in assignment (expression has type "Type[FakeResolverMatch]", variable has type "Optional[ResolverMatch]")
drf_source/tests/test_versioning.py:209: error: Incompatible types in assignment (expression has type "Type[FakeResolverMatch]", variable has type "Optional[ResolverMatch]")
drf_source/tests/test_versioning.py:260: error: Incompatible types in assignment (expression has type "Type[FakeResolverMatch]", variable has type "Optional[ResolverMatch]")
```

tests/typecheck/test_test.yml
```diff
@@ -26,3 +26,19 @@
...
+    response = client.get('/', format="json")
+    reveal_type(response) # N: Revealed type is "rest_framework.response._MonkeyPatchedResponse"
+
+- case: test_api_client_factory_get
+  main: |
+    from rest_framework import test
+
+    factory = test.APIRequestFactory()
+    request = factory.get("/")
+    reveal_type(request) # N: Revealed type is "rest_framework.request.Request"
```
intgr, line 44: Sorry, we don't ordinarily add tests for type hints, only for complex cases like decorators and overloads. Otherwise they would slow down the tests too much

intgr (commented):
Thanks for the PR

> Did not find a reason why it should not be there?

If there was a reason, I don't think it was a good reason. 

> some DRF test cases started to fail due to using fake implementations not inheriting original classes.

Yep, the typecheck is proving to be a bit of a PITA.

tests/typecheck/test_test.yml
piotrszyma: Should I revert changes to this file?

intgr: Yes please

piotrszyma: Reverted ✔️

intgr (approved):

PedroPerpetua:
Well then, good to merge and include in the next version? 😄

intgr:
There is a bit of work to do for the next version but yes 🥳

PR #335, opened 2023-02-02
Update Black to 23.1.0 and reformat with new style

intgr:
Black 23 introduced some changes to the stable formatting style, but very trivial in this project.

More details: https://github.com/psf/black/blob/main/CHANGES.md#2310

Closes #333

### DIFF
--- a/rest_framework-stubs/serializers.pyi
+++ b/rest_framework-stubs/serializers.pyi
@@ -1,16 +1,14 @@
 from collections.abc import Callable, Iterable, Iterator, Mapping, MutableMapping, Sequence
 from typing import Any, Generic, NoReturn, TypeVar
-from _typeshed import Self
 
+from _typeshed import Self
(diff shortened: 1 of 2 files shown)


rest_framework-stubs/serializers.pyi
```diff
@@ -1,16 +1,14 @@
...
 from django.core.exceptions import ValidationError as DjangoValidationError
 from django.db import models
 from django.db.models import DurationField as ModelDurationField
 from django.db.models import Manager, Model, QuerySet
 from django.db.models.fields import Field as DjangoModelField
-from django.utils.translation import ugettext_lazy as _
 from django.utils.functional import cached_property
-from typing_extensions import Literal
-
+from django.utils.translation import ugettext_lazy as _
```
intgr, line 11: isort also made changes to the import order, I thought lint was checking that already?

PR #336, opened 2023-02-02
Add max_length and min_length args for ListSerializer signature

imamatory:
`ListSerializer` signature missing a few arguments to specify validation of min and max length. See [docs](https://www.django-rest-framework.org/api-guide/serializers/#max_length)

https://github.com/encode/django-rest-framework/blob/22d206c1e0dbc03840c4d190f7eda537c0f2010a/rest_framework/serializers.py#L588-L589


intgr (approved):
Thanks, looks good.

I noticed that these can only be passed as keyword arguments. If you're up for it, would be nice to add `*,` separating positional and keyword-only arguments: https://peps.python.org/pep-3102/ Requires some reverse engineering in DRF code though.

PR #339, opened 2023-02-06
Upgrade to Mypy 1.0

christianbundy:
# I have made things!

Upgrade to Mypy 1.0

## Related issues

Refs https://github.com/python/mypy/issues/13685

Closes #340

### DIFF
--- a/scripts/typecheck_tests.py
+++ b/scripts/typecheck_tests.py
@@ -154,9 +154,9 @@ IGNORED_ERRORS = {
     "test_permissions.py": [
         '"ResolverMatch" has incompatible type "str"; expected "Callable[..., Any]"',
         "_SupportsHasPermission",
-        "Invalid type alias: expression is not a valid type",
-        '"object" not callable',
         'Cannot assign multiple types to name "composed_perm" without an explicit "Type[...]" annotation',
+        '"<typing special form>" not callable',
+        '<nothing> not callable',
     ],
     "test_relations.py": [
         'Invalid index type "int" for "Union[str, _StrPromise, List[Any], Dict[str, Any]]"; expected type "str"',
@@ -188,8 +188,6 @@ IGNORED_ERRORS = {
         "base class",
         '"CharField" has incompatible type "Collection[Any]"',
         'Name "foo" is not defined',
-        'Unsupported left operand type for | ("ReturnDict")',
-        'Unsupported left operand type for | ("Dict[str, str]")',
     ],
     "test_serializer_lists.py": [
         'Name "foo" is not defined',
@@ -208,7 +206,6 @@ IGNORED_ERRORS = {
     "test_testing.py": [
         '"Client" has no attribute "force_authenticate"',
         '"Client" has no attribute "credentials"',
-        'has no attribute "addClassCleanup"',
     ],
     "test_throttling.py": [
         'has incompatible type "Dict[<nothing>, <nothing>]"',
@@ -237,6 +234,7 @@ IGNORED_ERRORS = {
         "rest_framework.decorators",
         'Argument 1 to "include" has incompatible type "Tuple[List[object], str]"',
         'Incompatible types in assignment (expression has type "Type[FakeResolverMatch]", variable has type "Optional[ResolverMatch]"',  # noqa: E501
+        '"None" not callable'
     ],
     "test_viewsets.py": [
         '(expression has type "None", variable has type "Request")',
--- a/scripts/typecheck_tests.py
+++ b/scripts/typecheck_tests.py
@@ -154,6 +154,8 @@ IGNORED_ERRORS = {
     "test_permissions.py": [
         '"ResolverMatch" has incompatible type "str"; expected "Callable[..., Any]"',
         "_SupportsHasPermission",
+        "Invalid type alias: expression is not a valid type",
+        '"object" not callable',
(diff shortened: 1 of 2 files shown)


intgr:
Approved CI run.

intgr:
Some if not all of the typecheck errors are due to upstream changes in DRF test suite, not from upgrading mypy.

christianbundy:
Do we just add them as excluded errors in `./scripts/typecheck_tests.py`? Trying to make sure I understand the procedure.

intgr:
> Do we just add them as excluded errors in ./scripts/typecheck_tests.py?

I've looked at each error, and checked if it's something we're missing in stubs or a false positive. Usually it's a false positive, or some new feature that has been merged but will be in the next version (e.g. #413).

But it's annoying and confusing, especially for new contributors, when CI checks start to fail due to unrelated changes. So I've been thinking that we may want to freeze the git commit hash that we check (like is already done in django-stubs).

But the downside is that then we would have to periodically update this hash, something that's been a bit neglected in django-stubs.

intgr:
I rebased your branch. The remaining CI errors are from mypy 1.0 update.

christianbundy:
Thanks! For convenience, here are the relevant files:

- https://github.com/encode/django-rest-framework/blob/22d206c1e0dbc03840c4d190f7eda537c0f2010a/tests/test_permissions.py
- https://github.com/encode/django-rest-framework/blob/22d206c1e0dbc03840c4d190f7eda537c0f2010a/tests/test_versioning.py

## Problem 1

```
drf_source/tests/test_permissions.py:593: error: "<typing special form>" not callable
drf_source/tests/test_permissions.py:599: error: "<typing special form>" not callable
drf_source/tests/test_permissions.py:653: error: "<typing special form>" not callable
drf_source/tests/test_permissions.py:661: error: <nothing> not callable
drf_source/tests/test_permissions.py:673: error: "<typing special form>" not callable
drf_source/tests/test_permissions.py:681: error: <nothing> not callable
drf_source/tests/test_permissions.py:736: error: "<typing special form>" not callable
```

These seem to be exclusively complaining about this pattern:

```python3
composed_perm = permissions.IsAuthenticated | permissions.AllowAny
assert composed_perm().has_permission(request, None) is True
```

This only happens when you `__or__` permissions, and it looks like `__and__` is unaffected.

I don't really understand why this would happen, since [the stub definition](https://github.com/typeddjango/djangorestframework-stubs/blob/bf9d3837eaf0b1e2928f28bad48c1f617813376b/rest_framework-stubs/permissions.pyi#L18-L23) seems to treat `|` and `&` similarly, but I'm inclined to think that maybe this is a bug in Mypy where `|` is being treated as a type union. What do you think?

## Problem 2

```
drf_source/tests/test_versioning.py:28: error: "None" not callable
drf_source/tests/test_versioning.py:35: error: "None" not callable
drf_source/tests/test_versioning.py:43: error: "None" not callable
drf_source/tests/test_versioning.py:50: error: "None" not callable
```

These are all coming from:

```python3
self.versioning_class()
```

Where `self` is a subclass of `APIView` with a `get()` method defined. 

[The stub defines this as nullable](https://github.com/typeddjango/djangorestframework-stubs/blob/bf9d3837eaf0b1e2928f28bad48c1f617813376b/rest_framework-stubs/views.pyi#L54), and [DRF source](https://github.com/encode/django-rest-framework/blob/4abfa28e0879e2df45937ac8c7a9ffa161561955/rest_framework/views.py#L378) seems to agree that this is possible, so I can add an ignore for this one.

intgr:
> I don't really understand why this would happen, since [the stub definition](https://github.com/typeddjango/djangorestframework-stubs/blob/bf9d3837eaf0b1e2928f28bad48c1f617813376b/rest_framework-stubs/permissions.pyi#L18-L23) seems to treat | and & similarly, but I'm inclined to think that maybe this is a bug in Mypy where | is being treated as a type union. What do you think?

Yes indeed, mypy thinks `composed_perm` is a type alias for Union. For now, an ignore is fine.

I always hated how DRF requires permissions checks to be classes, not instances. 

This can cause problems in actual user code. Maybe report a mypy bug? Mypy could look at whether a class overrides the `__or__` method, or look at the method's return value, to detect this.

scripts/typecheck_tests.py
```diff
@@ -188,8 +188,6 @@
         "base class",
         '"CharField" has incompatible type "Collection[Any]"',
         'Name "foo" is not defined',
-        'Unsupported left operand type for | ("ReturnDict")',
-        'Unsupported left operand type for | ("Dict[str, str]")',
```
intgr, line 192: You can't remove these, they are still necessary for older Python versions.

intgr (changes requested):
There are still some test failures. Sorry, I could have approved the CI run earlier.

scripts/typecheck_tests.py
christianbundy: ah, I didn't realize that -- thanks!

christianbundy:
Done! And I ran `act` locally and it looks like it passed all tests. (Though I'd _love_ to be able to run CI without approvals if that's easy to fix.)

flaeppe:
> Done! And I ran `act` locally and it looks like it passed all tests. (Though I'd _love_ to be able to run CI without approvals if that's easy to fix.)

I've approved the CI run. You shouldn't need approvals once you've become a contributor.

intgr (approved):
Thanks!

PR #346, opened 2023-02-07
CI: Use django-stubs 1.14.0 for testing

intgr:
Needs changes to our typecheck ignores.

## Related issues

Closes #328

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,6 @@ djangorestframework==3.14.0
 types-pytz==2022.7.1.0
 types-requests==2.28.11.8
 types-urllib3==1.26.25.4
-django-stubs==1.13.0
+django-stubs==1.14.0
 django-stubs-ext==0.7.0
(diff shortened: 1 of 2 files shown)


mschoettle (approved):
It looks good to me.

I see there are also removals from the ignores. What is the process to update these when updating the stubs? Do they all have to be verified each time DRF or stubs are updated?

intgr:
> I see there are also removals from the ignores. What is the process to update these when updating the stubs?

At the end of typecheck run output there is a "UNUSED IGNORES" block, which is helpful for cleaning up unneeded ignores: https://github.com/typeddjango/djangorestframework-stubs/actions/runs/4132844223/jobs/7142029536#step:6:425

Although it doesn't currently work for `__common__` ignores. Also note that unused ignores vary by Python version, so some unused ones are expected.

intgr:
Thanks for the review!

PR #347, opened 2023-02-07
Run pre-commit isort on pyi files

intgr:
Pre-commit isort was ignoring `.pyi` files, this fixes that.

## Related issues

Similar to django-stubs PR https://github.com/typeddjango/django-stubs/pull/664


mschoettle (approved):
LGTM!

PR #359, opened 2023-02-22
Update package metadata: release notes URL, classifiers, maintainer

intgr:
Harmonized with django-stubs `setup.py`

* Added Release notes URL to package metadata (related: https://github.com/typeddjango/django-stubs/pull/365)
* Added classifiers for Python 3.11 and Django framework
* Added myself as the de-facto maintainer here
* Added type hints

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,9 +1,10 @@
 import os
+from typing import List
 
 from setuptools import find_packages, setup
 
 
-def find_stub_files(name):
+def find_stub_files(name: str) -> List[str]:
     result = []
     for root, _dirs, files in os.walk(name):
         for file in files:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -42,6 +43,8 @@ def find_stub_files(name):
     url="https://github.com/typeddjango/djangorestframework-stubs",
     author="Maksim Kurnikov",
     author_email="maxim.kurnikov@gmail.com",
+    maintainer="Marti Raudsepp",
+    maintainer_email="marti@juffo.org",
```
intgr, line 47: @sobolevn I noticed that maintainer was unset in djangorestframework-stubs `setup.py`. Should I update the maintainer field in both projects to myself?

OTOH if we ever go to a shared maintainership model, maybe it makes sense to remove it from both projects.

sobolevn: Please, feel free :)

It does not affect anything, but it is a nice way to say "thank you"!

sobolevn (approved):

PR #367, opened 2023-02-24
Allow user arg in `APIClient.force_authenticate` to be `None`

joshuadavidthomas:
[According to the Django REST Framework docs](https://www.django-rest-framework.org/api-guide/testing/#force_authenticateusernone-tokennone), the `user` argument can be set to `None` in the `force_authenticate` method on the `APIClient` class.


sobolevn (approved):
Thanks!

PR #376, opened 2023-03-16
Update `django-stubs` in `requirements.txt`

sobolevn:


intgr (approved):
:+1:

requirements.txt
```diff
@@ -7,6 +7,6 @@ djangorestframework==3.14.0
 types-pytz==2022.7.1.2
 types-requests==2.28.11.15
 types-urllib3==1.26.25.8
-git+https://github.com/typeddjango/django-stubs@update-mypy
+django-stubs==1.16.0
 django-stubs-ext==0.7.0
```
intgr, line 11: Oh we need to bump django-stubs-ext as well to avoid a dependency conflict.

```suggestion
django-stubs-ext==0.8.0
```

PR #379, opened 2023-03-21
fix: Update OrderingFilter get_default_ordering

albertocerrone:
### Why
`def get_default_ordering()` could return a None value in case ordering not be included in the view.

image


intgr (approved):
Thanks, looks correct.

PR #387, opened 2023-03-30
Improve APIException input and detail types

intgr:
APIException and its subclasses are more liberal in what they accept as input, but their `detail` attribute is already converted to a stricter type: translation strings are resolved, strings are conveted to ErrorDetail, etc

This uses the recursive types feature from mypy and thus bumps minimum mypy version to 0.991: https://mypy-lang.blogspot.com/2022/11/mypy-0990-released.html


intgr:
I think this is complete. Reviews welcome.

sobolevn (approved):
The usage seems correct to me, but I am not an expert in this API :(

PR #414, opened 2023-05-12
replace OrderedDict with dict

terencehonles:
# I have made things!

This change replaces `OrderedDict` with `dict` as done in https://github.com/encode/django-rest-framework/pull/8964. This does not look like it has been released yet, but since this package requires Python 3.8 which is > Python 3.5, which was the last release to need `OrderedDict`, then this change should be safe to merge now. It should also be safe because `dict` is a superclass of `OrderedDict` and should not change anything for users who are using `OrderedDict` still.


sobolevn (approved):

sobolevn:
Thanks!

PR #415, opened 2023-05-12
Changed from list to abstract Sequence type for http verbs

namper:
Allows method decorator to not fail on tuple, as there is no reason to have `methods` mutable

# I have made things!
Changed from `list` to abstract `Sequence` type for http verbs. while documentation uses `list` for declaring methods for action decorator, the actual implementation only uses fact that methods are `Iterable`, and they rebuild this list all over again for normalisation. 

Documentation: https://www.django-rest-framework.org/api-guide/viewsets/#viewset-actions


sobolevn (approved):
Thanks!

PR #417, opened 2023-05-13
CI: Always install django-stubs from git master

intgr:
It's not uncommon to need to coordinate changes between django-stubs and djangorestframework-stubs projects. E.g. I need to make some changes in django-stubs first and then also make a similar change here that builds on django-stubs. Happens particularly often with mypy updates like #416.

So I propose making django-stubs permanently a git dependency in CI and development `requirements.txt`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,6 @@ djangorestframework==3.14.0
 types-pytz==2023.3.0.0
 types-requests==2.30.0.0
 types-urllib3==1.26.25.13
-django-stubs==4.2.0
+git+https://github.com/typeddjango/django-stubs
 django-stubs-ext==4.2.0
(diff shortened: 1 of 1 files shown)


intgr:
A bit offtopic but related: Most of the problems with using older `django-stubs` are failures in our CI that do not occur in actual user projects (in my experience).

But there's a risk of creating a dependency on other changes already merged in `django-stubs` and we don't test against multiple django-stubs versions. So I'm thinking going forward we should consider bumping DRF-stubs dependency on django-stubs with every DRF-stubs release.

sobolevn (commented):
Yes, seems reasonable. We already synchronize releases, so it can decrease the amount of problems during development with no real risks.

PR #425, opened 2023-06-06
CI: Add suppressions due to django-stubs changes

intgr:
Fixes new errors that appeared after django-stubs changes.

```
drf_source/tests/test_response.py:259: note: "Dict" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
drf_source/tests/test_response.py:259: note: Consider using "Mapping" instead, which is covariant in the value type
drf_source/tests/test_response.py:269: note: "Dict" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
drf_source/tests/test_response.py:269: note: Consider using "Mapping" instead, which is covariant in the value type
drf_source/tests/test_response.py:278: note: "Dict" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
drf_source/tests/test_response.py:278: note: Consider using "Mapping" instead, which is covariant in the value type
```

### DIFF
--- a/scripts/typecheck_tests.py
+++ b/scripts/typecheck_tests.py
@@ -188,6 +188,8 @@ IGNORED_ERRORS = {
     ],
     "test_response.py": [
         'Argument 2 to "get" of "Client" has incompatible type "**Dict[str, str]"',
+        '"Dict" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance',
+        'Consider using "Mapping" instead, which is covariant in the value type',
     ],
     "test_routers.py": [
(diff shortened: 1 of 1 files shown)


sobolevn (commented):
I think that we can use `Mapping` there, the only method that is using `headers` is https://github.com/django/django/blob/4f6a51dfe6a4a8d5ed61c73b902e808109f142b0/django/http/request.py#L461

And it allows `Mapping`.

intgr:
Fixed in https://github.com/typeddjango/django-stubs/pull/1537 instead, closing.

PR #439, opened 2023-06-27
CI: Remove typecheck tests and clean things up

intgr:
* Removed everything related to `typecheck_tests.py`
* Made true/false values lowercase in `mypy.ini` like django-stubs

## Related issues

* Like django-stubs: https://github.com/typeddjango/django-stubs/pull/1556


sobolevn (approved):
Thank you!

PR #441, opened 2023-06-27
Convert _typeshed.Self usages to typing_extensions.Self (PEP 673)

intgr:
`typing_extensions.Self` is easier to use, since it no longer requires hinting the `self` or `cls` parameter, the self type is inferred automatically. Docs: https://peps.python.org/pep-0673/

Using Ruff `flake8-tidy-imports` to disallow reintroducing this import in the future.

## Related issues

Similar to https://github.com/typeddjango/django-stubs/pull/1591


sobolevn (approved):
👍

Issue #255, opened 2022-09-21
GenericAPIView has missing type parameters, but when added 'type' object is not subscriptable


XF-FW:
# Bug report
## What's wrong

When using mypy with strict=true, it issues the following error:
`Missing type parameters for generic type "GenericAPIView"`. 

So, I check the type and add it, but I'm now getting a runtime error:
`TypeError: 'type' object is not subscriptable`.

I am using `django_stubs_ext.monkeypatch()`, which solves the problem for other similar issues.

I haven't test it on its own, but this should suffice, as a reproducible test:
```
from rest_framework.generics import GenericAPIView

class ShouldWorkView(GenericAPIView[None]):
   ...
```

## How is that should be

The above example should pass, with no issues.

## System information

- OS: Arch
- `python` version: 3.10
- `django` version: 4.0.7
- `mypy` version: 0.942
- `django-stubs` version: 1.12.0

ewiner:
I believe you're hitting this issue: https://mypy.readthedocs.io/en/latest/runtime_troubles.html#using-classes-that-are-generic-in-stubs-but-not-at-runtime.  The solution suggested on that doc page worked for me.

XF-FW:
Yep, that worked. Thanks!

It still feels like a workaround, so I'll leave this open. If the maintainers feel like this should be the default behavior, feel free to close it.

XF-FW:
The same thing happens when trying to subclass `PrimaryKeyRelatedField`

monosans:
You can pass the classes you want to monkeypatch in the `extra_classes` parameter to `django_stubs_ext.monkeypatch()`.

Example:

```python
import django_stubs_ext
from rest_framework.generics import GenericAPIView

django_stubs_ext.monkeypatch(extra_classes=(GenericAPIView,))
```

or you can manually monkeypatch them:

```python
from rest_framework.generics import GenericAPIView

GenericAPIView.__class_getitem__ = (  # type: ignore[attr-defined]
    classmethod(lambda cls, *args, **kwargs: cls)
)
```

XF-FW:
Thank you! That works perfectly.

Issue #164, opened 2021-08-16
Missing type parameters for generic type “ModelSerializer”


sant527:
# Bug report

I have 

```
class AnimalSerializer(serializers.ModelSerializer):
    class Meta:
        model = Animal
        fields = "__all__"
```

Now i run mypy 

```
[mypy]
# The mypy configurations: https://mypy.readthedocs.io/en/latest/config_file.html
python_version = 3.9

check_untyped_defs = True
# disallow_any_explicit = True
disallow_any_generics = True
disallow_untyped_calls = True
disallow_untyped_decorators = True
ignore_errors = False
ignore_missing_imports = True
implicit_reexport = False
strict_optional = True
strict_equality = True
no_implicit_optional = True
warn_unused_ignores = True
warn_redundant_casts = True
warn_unused_configs = True
warn_unreachable = True
warn_no_return = True
mypy_path = /home/simha/app/backend_django/src

plugins =
  mypy_django_plugin.main,
  mypy_drf_plugin.main

[mypy.plugins.django-stubs]
django_settings_module = petsproject.settings
```
```
(venv) $ mypy .
```

I get 

```
error: Missing type parameters for generic type "ModelSerializer"

```

## System information

- OS:
- `python` 3.9:
- `django` 3.1:
- `mypy` 0.910
- `django-stubs` 1.8.0
- `djangorestframework-stubs` 1.4.0

sobolevn:
This is how `ModelSerializer` is defined: https://github.com/typeddjango/djangorestframework-stubs/blob/master/rest_framework-stubs/serializers.pyi#L214

You can:
1. Define your subtype as `class AnimalSerializer(serializers.ModelSerializer[Animal]):`
2. Disable these errors with `disallow_any_generics = False`

I hope this helps!

browniebroke:
> 1. Define your subtype as `class AnimalSerializer(serializers.ModelSerializer[Animal]):`

I didn't realise that was supported until stumbling upon this issue, this is amazing! Did I miss the documentation or it doesn't exist? I noticed that the `django-stubs` project has a "Features" section showing these kind of things, would it make sense to create one on this project too? I'm happy to send a PR

Issue #299, opened 2022-11-22
Monkeypatching breaks 


jerivas:
# Bug report

## What's wrong

It appears that mokeypatching DRF viewsets (as suggested [here](https://github.com/typeddjango/djangorestframework-stubs/issues/255#issuecomment-1320496964)) breaks the usage of `django_filters` FilterSets:

```python
# views.py
class DocumentViewSet(viewsets.ModelViewSet[Document]):
    queryset = Document.objects.all()
    serializer_class = DocumentSerializer
    filterset_class = DocumentFilterSet

# settings.py
import django_stubs_ext
from rest_framework import viewsets

django_stubs_ext.monkeypatch(extra_classes=(viewsets.ModelViewSet,))
```
With this code the `DocumentViewSet` completely ignores the filters defined in `DocumentFilterSet`. GET params simply stop working for filtering and the filter UI is completely missing from DRF's HTML view.

_Refer to this [test](https://github.com/jerivas/filter-bug-repro/blob/58cd0649617e9d6b75f903e256d1a364af3b4111/filter_bug_repro/tests.py) in the demo repository. In the [original commit](https://github.com/jerivas/filter-bug-repro/blob/01c92a55a0f1f1e096d3d3a42c6f17ccdaa63057/filter_bug_repro/urls.py), the test passes, but after [types are added](https://github.com/jerivas/filter-bug-repro/commit/58cd0649617e9d6b75f903e256d1a364af3b4111), the test fails._

## How is that should be

DRF filters should continue working after monkeypatching.

## System information

- OS: Ubuntu 20.04
- `python` version: 3.10
- `django` version: 4.1
- `mypy` version: 0.991
- `django-stubs` version: 1.13.0
- `django-stubs-ext` version: 0.7.0

LucidDan:
This happened to me this week on a production app for a client. We realised the problem because our Zapier integration stopped working when the REST API no longer had search and pagination support. Took me a whole day to track the problem down.

### Root Cause
I think what is actually happening here is that importing from rest_framework in the settings.py causes django-rest-framework to fail to load any settings and just use its defaults. Unfortunately, it does this silently, so you can easily end up with things breaking without realising it.

I think this is not really a djangorestframework-stubs bug. It's maybe not even really a rest-framework "bug" - more of a feature of past design choices (eg choosing to allow REST_FRAMEWORK to be optional and not throw an error if it isn't defined).

### Workarounds
[...]

PR #444, opened 2023-07-10
Add FAQ section in README to explain usage of Generics

browniebroke:
# I have made things!

I recently came across the fact that this library adds support for Generics in DRF serializers, which makes type checking a lot better. It might be obvious to the people who are familiar with type checking, but it wasn't obvious to me that this library supported that.

## Related issues

Refs #164

## Follow up

I think it's also supported for GenericViewSet, but requires [a bit more extra setup](https://github.com/typeddjango/djangorestframework-stubs/issues/255). If that's correct, I can also add a paragraph/section about that, although I'd like to get feedback on this change first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,20 @@ plugins =
 
 in your `mypy.ini` file.
 
+## FAQ
+
+### Model instance is inferred as `Any` instead of my `Model` class
+
+This library adds support for [Generics](https://peps.python.org/pep-0484/#generics) in DRF. For example, you can now write:
+
+```python
+class MyModelSerializer(serializers.ModelSerializer[MyModel]):
+    class Meta:
+        model = MyModel
+        fields = ("id", "example")
+```
+
+Which means that methods where the model is being passed around will know the actual type of the model instead of being `Any`. The `instance` attribute on the above serializer will be `Union[MyModel, typing.Sequence[MyModel], None]`.
 
 ## To get help
 


README.md
```diff
@@ -26,6 +26,20 @@ plugins =
...
 in your `mypy.ini` file.
 
+## FAQ
+
+### Model instance is inferred as `Any` instead of my `Model` class
+
+This library adds support for [Generics](https://peps.python.org/pep-0484/#generics) in DRF. For example, you can now write:
+
+```python
+class MyModelSerializer(serializers.ModelSerializer[MyModel]):
```
sobolevn, line 36: Do we need to monkeypatch it?

browniebroke: Do you mean call `django_stubs_ext.monkeypatch()`? I'm pretty sure it's not needed, worked without it in my project.

browniebroke: The `__class_getitem__` method was added https://github.com/encode/django-rest-framework/pull/7385 which was released in DRF 3.12

browniebroke:
> I think it's also supported for GenericViewSet, but requires [a bit more extra setup](https://github.com/typeddjango/djangorestframework-stubs/issues/255).

I tried to use it in my project but it didn't work. After a quick search, the monkeypatching might not be working [so well for DRF](https://github.com/typeddjango/djangorestframework-stubs/issues/299#issuecomment-1343952084), depending on your DRF settings.

From what I understand, it's not properly fixable in any of the projects from the @typeddjango org, and it should be better once DRF 3.15 is out, which will include [this change](https://github.com/encode/django-rest-framework/pull/8825). I suggest to not advertise this feature for views until 3.15 is out, and it will need a note about the minimum version.

sobolevn:
Do we have a ETA for the release?

browniebroke:
There is a [PR open](https://github.com/encode/django-rest-framework/pull/8794) to prepare the release notes. It's been open for a while now and it looks like it's all ready. Probably just waiting on a maintainer to find enough free time to do it.

README.md
```diff
@@ -26,6 +26,20 @@ plugins =
 
 in your `mypy.ini` file.
 
+## FAQ
+
+### Model instance is inferred as `Any` instead of my `Model` class
+
+This library adds support for [Generics](https://peps.python.org/pep-0484/#generics) in DRF. For example, you can now write:
```
intgr, line 33: I would rephrase this:

```suggestion
When subclassing `ModelSerializer`, add a [type argument](https://peps.python.org/pep-0484/#generics) to type-hint the related model class, for example:
```

Feel free to tweak this further.

PR #452, opened 2023-07-25
Fix ReturnDict and ReturnList __init__ signatures

### LINKED ISSUE #217: ReturnDict annotations incorrect
# Bug report

## What's wrong

The stubs say that `ReturnDict` has `serializer` as first argument, but actually it's a keyword only argument. First argument should be a Mapping. (or whatever parent `OrderedDict` takes).

```
from rest_framework.utils.serializer_helpers import ReturnDict
from rest_framework.serializers import Serializer

d = ReturnDict({"key": "val"}, serializer=Serializer())
```

produces:

```
test.py:5: error: "ReturnDict" gets multiple values for keyword argument "serializer"  [misc]
test.py:5: error: Argument 1 to "ReturnDict" has incompatible type "Dict[str, str]"; expected "BaseSerializer[Any]"  [arg-type]
```

## How is that should be

It should not give an error. That's the usage in drf itself

https://github.com/encode/django-rest-framework/blob/3.13.1/rest_framework/utils/serializer_helpers.py#L16
https://github.com/encode/django-rest-framework/blob/3.13.1/rest_framework/serializers.py#L553

```
@property
def data(self):
    ret = super().data
    return ReturnDict(ret, serializer=self)
```

## System information

- `django-stubs` version: 1.5.0
- `django-rest-framework` version: 3.13.1

moosbruggerj:
# Fix `__init__` signatures of ReturnDict and ReturnList

PR makes `serializer` argument in `__init__` method of `ReturnDict` and `ReturnList` position only and otherwise includes `list` and `dict` type annotations from typeshed.
## Related issues

- Closes #217

### DIFF
--- a/rest_framework-stubs/utils/serializer_helpers.pyi
+++ b/rest_framework-stubs/utils/serializer_helpers.pyi
@@ -1,5 +1,5 @@
 from collections.abc import Iterable, Iterator, MutableMapping
-from typing import Any, TypeVar, overload
+from typing import Any, Generic, TypeVar, overload
 
 from _typeshed import SupportsKeysAndGetItem
 from rest_framework.exceptions import ErrorDetail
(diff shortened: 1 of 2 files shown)


rest_framework-stubs/utils/serializer_helpers.pyi
```diff
@@ -1,19 +1,51 @@
...
+from _typeshed import SupportsKeysAndGetItem
 from rest_framework.exceptions import ErrorDetail
 from rest_framework.fields import Field
 from rest_framework.serializers import BaseSerializer
 
+_T = TypeVar("_T")
+_VT = TypeVar("_VT")
+_KT = TypeVar("_KT")
+
 class ReturnDict(dict):
```
sobolevn, line 13: Shouldn't it be generic?

rest_framework-stubs/utils/serializer_helpers.pyi
```diff
@@ -1,19 +1,51 @@
...
+        self: dict[str, _VT], __iterable: Iterable[tuple[str, _VT]], *, serializer: BaseSerializer, **kwargs: _VT
+    ) -> None: ...
+    # Next two overloads are for dict(string.split(sep) for string in iterable)
+    # Cannot be Iterable[Sequence[_T]] or otherwise dict(["foo", "bar", "baz"]) is not an error
+    @overload
+    def __init__(self: dict[str, str], __iterable: Iterable[list[str]], *, serializer: BaseSerializer) -> None: ...
+    @overload
+    def __init__(
+        self: dict[bytes, bytes], __iterable: Iterable[list[bytes]], *, serializer: BaseSerializer
+    ) -> None: ...
```
intgr, line 39: Wow these overloads are a lot more complicated than I imagined.

I'm not sure if `ReturnDict` makes sense with non-string keys. Only string keys are JSON-serializable. But I'm OK with keeping this as is.

rest_framework-stubs/utils/serializer_helpers.pyi
intgr: I think it makes sense to make key/value generic, like dict, since it subclasses dict. Don't bother with generic serializer, unless there's a particular reason to do so.

rest_framework-stubs/utils/serializer_helpers.pyi
```diff
@@ -1,19 +1,51 @@
...
 from rest_framework.serializers import BaseSerializer
 
+_T = TypeVar("_T")
+_VT = TypeVar("_VT")
+_KT = TypeVar("_KT")
+
 class ReturnDict(dict):
     serializer: BaseSerializer
-    def __init__(self, serializer: BaseSerializer = ..., *args, **kwargs): ...
+    # def __init__(self, serializer: BaseSerializer, *args, **kw    @overload
```
intgr, line 15: Let's leave a comment about the origin of these.

```suggestion
    # Copied from https://github.com/python/typeshed/blob/main/stdlib/builtins.pyi `class dict`
```

intgr (approved):
This is great, thanks!

intgr:
Finally released in version 3.14.3

PR #464, opened 2023-09-12
Annotate Field.default_error_messages as a ClassVar

brianhelba:
This clarifies that this field is set on the class itself, not on instances.


brianhelba:
@sobolevn The tests on `master` appear to be broken. I don't think this is a failure with this PR.

brianhelba:
Tests are fixed by #465.

PR #465, opened 2023-09-12
Fix broken tests caused by django-stubs 4.2.4

brianhelba:


brianhelba:
@sobolevn PTAL.

tests/typecheck/test_views.yml
```diff
@@ -16,7 +16,7 @@
         pass
 
     class MyView(generics.DestroyAPIView):
-        queryset = MyModel.objects.all()
+        queryset = MyModel._default_manager.all()
```
sobolevn, line 19:
```suggestion
        queryset = MyModel.objects.all()
```

tests/typecheck/test_views.yml
```diff
@@ -16,7 +16,7 @@
         pass
```
sobolevn, line 16:
```suggestion
        field = models.IntegerField()
```

sobolevn (commented):
This should also fix the issue

brianhelba:
Ah, makes sense. Applied your suggestion.

brianhelba:
@sobolevn I thought I understood your suggestion: since `MyModel` had no fields, it was generic and had no `.objects`.

However, it looks like the tests are actually still failing with your suggestion.

sobolevn (approved):

sobolevn:
Thank you!

brianhelba:
No, the tests are still failing. I just pushed the import fix to isolate the actual breakage:
> main:8: error: "Type[MyModel]" has no attribute "objects"

sobolevn:
This looks like a bug

intgr:
I have merged @christianbundy's fix in PR https://github.com/typeddjango/djangorestframework-stubs/pull/471.

PR #469, opened 2023-09-21
Specify return type for APIClient.credentials method

ltakens:
Help type checkers infer the return type of the `.credentials` method on the `APIClient` class.

E.g. vscode's **Pylance** extension seems to report a partially unknown type:

Screenshot 2023-09-21 at 23 41 42

**mypy** and **pyright** don't report this error when run from the command line.


intgr (approved):
Thanks!

The test failures are unrelated and due to `django-stubs` changes. I can hopefully fix them tomorrow, and then also merge this PR.

PR #471, opened 2023-09-25
Fix test broken by django-stubs 4.2.4

christianbundy:
**Problem**

The new version of Django stubs broke a test.

**Solution**

Fix it. It looks like we need to expose the model in `models` for it to get `.objects`.


intgr (approved):
Thanks for the help! Sorry, I've neglected this project recently.

PR #476, opened 2023-10-04
Specify that Request.parser_context can never be None

### LINKED ISSUE #473: Request.parser_context is typed as possibly None while it never can be None
## What's wrong

`rest_framework.request.Request.parser_context` is typed as possibly being `None`. However according to the constructor of the `Request` class it can never actually be `None` because if it is it would be set to an empty dict instead, see https://github.com/encode/django-rest-framework/blob/master/rest_framework/request.py#L171.
Due to this wrong typing I now either have to code for an impossible scenario or disable a mypy rule, both I rather not do.

## How is that should be

It should be typed as just `dict[str, Any]` rather than `dict[str, Any] | None`.

## System information

- OS: Alpine Linux
- `python` version: 3.11.5
- `django` version: 4.2
- `mypy` version: 1.5
- `django-stubs` version: 3.14.2

intgr:
Closes #473.

### DIFF
--- a/rest_framework-stubs/request.pyi
+++ b/rest_framework-stubs/request.pyi
@@ -45,7 +45,7 @@ class Request(HttpRequest):
     parsers: Sequence[BaseParser] | None
     authenticators: Sequence[BaseAuthentication | ForcedAuthentication] | None
     negotiator: BaseContentNegotiation | None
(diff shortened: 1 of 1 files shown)


intgr:
Merging myself as trivial change.

PR #479, opened 2023-10-04
Harmonize `mypy.ini` config with django-stubs

intgr:
* Most of the overrides in `tests/plugins.ini` are no longer needed
* Removed lots of unneeded `# type: ignore` comments
* Will create a separate PR to update `disallow_untyped_defs` and `disallow_incomplete_defs` too

## Related issues

* https://github.com/typeddjango/django-stubs/pull/663

PR #480, opened 2023-10-04
Fill remaining missing hints & disallow partial hints

intgr:
Disallow untyped/incomplete function definitions (`disallow_untyped_defs = true`, `disallow_incomplete_defs = true`)

But using `_typeshed.Incomplete` instead of `Any` as the placeholder type.

## Related issues

* https://github.com/typeddjango/django-stubs/pull/1206
* Depends on #479


mypy.ini
```diff
@@ -27,3 +26,19 @@ django_settings_module = scripts.drf_tests_settings
 
 [mypy-uritemplate.*]
 warn_unreachable = false
+
+[mypy-yaml.*]
+disallow_untyped_defs = false
+disallow_incomplete_defs = false
```
intgr, line 32: I think these are necessary due to the way `pytest-mypy-plugins` works.

Mypy ordinarily ignores errors from site-packages imports, but they aren't ignored in tests.

sobolevn (approved):
Thanks! I skimmed through, this is not a detailed review :)
We can fix any annotation later if needed.

mypy.ini
sobolevn: This can be a new bug / feature request in `pytest-mypy-plugins`. Users should not see this.

intgr: 👍 Reported issue https://github.com/typeddjango/pytest-mypy-plugins/issues/134

PR #485, opened 2023-10-12
Allow `None` for `HyperlinkedRelatedField.get_url(format=)` parameter

4c0n:
# I have made things!

Hi! I ran into a false positive when running mypy in combination with this package.
Actually the `format` parameter can also be `None` in the `get_url()` method, it is passed to `reverse()` where the default value also is `None`. We've been using it like this in our application for a long time and it works fine like that.
Please see:
https://github.com/encode/django-rest-framework/blob/3.14.0/rest_framework/relations.py#L326
https://github.com/encode/django-rest-framework/blob/3.14.0/rest_framework/reverse.py#L32

Thanks in advance for your time and consideration.

## Related issues
I didn't create an issue, should I?


intgr (approved):
Thanks!

> I didn't create an issue, should I?

No need. Only for more complex changes that you think warrant prior discussion.

Issue #434, opened 2023-06-22
Use `stubtest` for DRF as well


sobolevn:
It can be done similar to https://github.com/typeddjango/django-stubs/pull/1560

PR #491, opened 2023-10-18
Update Field constructor kwargs-only parameters

### LINKED ISSUE #450: update serializers Field arguments to be named-only
# Bug report

basically apply the changes https://github.com/encode/django-rest-framework/pull/7632 but to the stubs

## What's wrong

I'm attempting to upgrade drf -- and I'd have a lot more confidence if the type checker could help me find the places I'm missing

## How is that should be

see above

## System information

- OS:
- `python` version: n/a
- `django` version: n/a
- `mypy` version: n/a
- `django-stubs` version: n/a

intgr:
# I have made things!

The number of allowed positional parameters of `Field` class constructors was significantly reduced by DRF upstream, but some remain regardless.

Also correctly reordered arguments where positional args _are allowed_.

## Related issues

* Fixes #450
* Refs https://github.com/encode/django-rest-framework/pull/7632


sobolevn (commented):
I think we need a `stubtest` for this :(

intgr:
Agreed, stubtest would help for sure and it's on my radar. I've assigned #434 to myself.

scripts/stubtest/allowlist_todo.txt
```diff
@@ -46,28 +46,20 @@ rest_framework.fields.DateField.to_internal_value
 rest_framework.fields.DateTimeField.__init__
 rest_framework.fields.DateTimeField.to_internal_value
 rest_framework.fields.DecimalField.__init__
-rest_framework.fields.DictField.__init__
 rest_framework.fields.DictField.initial
 rest_framework.fields.DurationField.to_internal_value
 rest_framework.fields.Field.__init__
 rest_framework.fields.FilePathField.__init__
```
intgr, line 52: Note: some `XxxField.__init__` methods remain because argument types aren't precise. But I'd rather fix that separately.

sobolevn (approved):
stubtest gives a lot of confidence! Thanks!

PR #492, opened 2023-10-18
Cleanup: Add missing `None` return to `__init__` methods

intgr:
Seemed a bit inconsistent. This has no effect on user code.

## Related issues

* Follow-up to #480


intgr:
Merging myself as trvial change.

PR #494, opened 2023-10-18
Version 3.14.4 release

intgr:
Tests will be failing until `django-stubs` PR is merged.

* Depends on https://github.com/typeddjango/django-stubs/pull/1777
* Related https://github.com/typeddjango/django-stubs/issues/1740

PR #495, opened 2023-10-19
Drop hard dependency on mypy

intgr:
As [announced in 3.14.4 release notes](https://github.com/typeddjango/djangorestframework-stubs/releases/tag/3.14.4), we will drop the hard dependency on mypy. Users of djangorestframework-stubs with mypy will need to add their own dependency on mypy, or use `djangorestframework-stubs[compatible-mypy]` extra.

Currently mypy remains the only supported type checker. Improvements for other type checkers may be considered in the future, pull requests welcome.

## Related issues

* Related: https://github.com/typeddjango/django-stubs/pull/1782
* Related: https://github.com/typeddjango/django-stubs/issues/1628

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,6 @@ with open("README.md") as f:
     readme = f.read()
 
 dependencies = [
-    "mypy>=0.991",
     "django-stubs>=4.2.5",
     "typing-extensions>=3.10.0",
     "requests>=2.0.0",
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ dependencies = [
 
 # Keep compatible-mypy major.minor version pinned to what we use in CI (requirements.txt)
 extras_require = {
-    "compatible-mypy": ["mypy~=1.6.0"],
+    "compatible-mypy": ["mypy~=1.6.0", "django-stubs[compatible-mypy]"],
     "coreapi": ["coreapi>=2.0.0"],
     "markdown": ["types-Markdown>=0.1.5"],
 }
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,6 @@ djangorestframework==3.14.0
 types-pytz==2023.3.1.1
 types-requests==2.31.0.10
 types-urllib3==1.26.25.14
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -20,7 +20,6 @@ def find_stub_files(name: str) -> List[str]:
     readme = f.read()
 
 dependencies = [
-    "mypy>=0.991",
     "django-stubs>=4.2.5",
```
sobolevn, line 23: We also have to bump `django-stubs` here.

Maybe we should use `django-stubs[compatible-mypy]` here? 
This way we can be sure that the version resolution works correctly for both packages.

intgr: > We also have to bump django-stubs here.

Yep, I bump this dependency every time I make a release, e.g. https://github.com/typeddjango/djangorestframework-stubs/pull/494/files

> Maybe we should use django-stubs[compatible-mypy] here?

Nope, that would re-introduce the mypy dependency.

intgr: Hmm, we *could* add "django-stubs[compatible-mypy]" to the extras deps, e.g.

```python
extras_require = {
    "compatible-mypy": ["mypy~=1.6.0", "django-stubs[compatible-mypy]"],
```

sobolevn: yes, this is what I meant :+1:

intgr: It has happened a few times that `django-stubs` and `djangorestframework-stubs` versions for `compatible-mypy` get out of sync.

My understanding is that pip's constraint resolver is pretty primitive. Not totally sure, but I'm afraid pip would just give up in such a situation, and make `djangorestframework-stubs[compatible-mypy]` uninstallable.

Testing this hypothesis is difficult though. So maybe err on the side of caution and not introduce this?

However, what does make sense is adding `django-stubs[compatible-mypy]` dependency to `requirements.txt`, so our CI will alert such an issue.

intgr: I played around with this and seems like my fears were unfounded. When changing the dependencies around,  `pip` was successfully capable of finding the newest `django-stubs` with a compatble mypy dependency.

Let's go with that.

sobolevn (approved):
Thanks, this will be helpful for poetry and pdm users!

PR #488, opened 2023-10-17
Migrate from flake8 to ruff

intgr:
## Related issues

Similar to change in django-stubs: https://github.com/typeddjango/django-stubs/pull/1718. Thanks to @GabDug for the original PR there.

PR #496, opened 2023-10-22
Fix and enable Ruff lints PYI002, PYI026

intgr:
These affect `rest_framework-stubs/compat.pyi` which is using some weird hacks that Ruff and mypy don't really like.

* PYI002: If test must be a simple comparison against sys.platform or sys.version_info
* PYI026: Use typing_extensions.TypeAlias for type aliases

## Related issues

These were deferred from #488.


rest_framework-stubs/compat.pyi
```diff
@@ -1,51 +1,56 @@
...
 
 except ImportError:
-    apply_markdown = None  # type: ignore
-    markdown = None  # type: ignore
+    apply_markdown: TypeAlias = None  # type: ignore[no-redef]
+    markdown: TypeAlias = None  # type: ignore[no-redef]
 
-if markdown is not None and pygments is not None:
+try:
+    import pygments
```
sobolevn, line 43: Why do we need `import pygments` here? I seems to be not used and not re-exported (if this was the goal) 🤔

sobolevn: We also have it here: https://github.com/typeddjango/djangorestframework-stubs/pull/496/files#diff-eca906d3868175be921f64f15c140de74981da18a478ae18da601a43a969deb8R31

intgr: It is re-exported in `__all__`

```python
__all__ = [
    ...
    "pygments",
```

intgr: I've simplified this, merged it into the `try: import markdown` block.

sobolevn (approved):

PR #497, opened 2023-10-22
Use explicit mypy error codes in ignore comments and output

intgr:
* Adds mypy error codes to output (affects tests)
* Requires explicit error codes in `# type: ignore` comments, e.g. `type: ignore[assignment]`

## Related issues

* Follow-up to #496


intgr:
Draft because #496 should be merged first.

PR #499, opened 2023-10-22
Add all stubs reported missing by stubtest

intgr:
I filtered all stubtest reports that have message "is not present in stub" or "failed to find stubs" and fixed them.


sobolevn (approved):
Great work!

PR #500, opened 2023-10-23
change get_permissions to return list

mschoettle:
# I have made things!

Proposal to change the signature of `get_permissions` to `list[BasePermission]`. This allows one to customize the permissions of a view as suggested by the documentation: https://www.django-rest-framework.org/api-guide/permissions/#overview-of-access-restriction-methods

The source code also suggests that is is a `list`:

https://github.com/encode/django-rest-framework/blob/f56b85b7dd7e4f786e0769bba6b7609d4507da83/rest_framework/views.py#L278

It was originally changed in #320 to `Sequence[_SupportsHasPermission]`. I am not sure if the element type has to be `_SupportsHasPermission` instead of `BasePermission`. All the other method signatures use the base type.

## Related issues

Refs #319 (PR: #320)


mschoettle:
I found a case that supports `_SupportsHasPermission` instead of `BasePermission`:

```py
def get_permissions(self) -> Sequence[_SupportsHasPermission]:
        if self.action == 'list':
            return [IsAdminUser | SomeOtherPermission]

        return super().get_permissions()
```

produces

> List item 0 has incompatible type "OperandHolder"; expected "_SupportsHasPermission"  [list-item]

mschoettle:
I had to revert back to `Sequence`. Given that in some scenarios (see added test cases) `_SupportsHasPermission` has to be exposed I think it would be best to make it part of the public API. Thoughts?

PR #504, opened 2023-11-02
Improve `Serializer.to_representation()` return type

4c0n:
This PR adds a more specific return type to `to_representation()` .
We can make up from the documentation and the implementation of `Serializer` that a dictionary is to be returned.

Please see:
https://www.django-rest-framework.org/api-guide/serializers/#baseserializer
https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L522

Thanks for your time and consideration!


rest_framework-stubs/serializers.pyi
```diff
@@ -225,7 +225,7 @@ class ModelSerializer(Serializer, BaseSerializer[_MT]):
     def update(self, instance: _MT, validated_data: Any) -> _MT: ...
     def create(self, validated_data: Any) -> _MT: ...
     def save(self, **kwargs: Any) -> _MT: ...
-    def to_representation(self, instance: _MT) -> Any: ...
+    def to_representation(self, instance: _MT) -> dict[str, Any]: ...
```
intgr, line 228: Looks like `ModelSerializer` doesn't override this method, we can simply remove and it will be inherited.
```suggestion
```

intgr (approved):
Thanks!

PR #508, opened 2023-11-15
Replace `Manager` with `BaseManager`

intgr:
@flaeppe This is some example fallout from https://github.com/typeddjango/django-stubs/pull/1814 since `RelatedManager` base class was changed from `Manager` to `BaseManager`.

Without this change, I can no longer pass `RelatedManager`s to APIs that accept them.

I'm not entirely happy or sure about this. I think in pratice most managers, including `RelatedManager`s, do inherit from `Manager`.

`BaseManager` seems to be intended as an impementation detail in Django, not something users should be exposed to. But this will force users to replace `Manager` type hints with `BaseManager`, including my personal projects.

## Related issues

* https://github.com/typeddjango/django-stubs/pull/1814

### DIFF
--- a/rest_framework-stubs/generics.pyi
+++ b/rest_framework-stubs/generics.pyi
@@ -2,7 +2,8 @@ from _typeshed import Incomplete
 from collections.abc import Sequence
 from typing import Any, Protocol, TypeVar
 
-from django.db.models import Manager, Model
+from django.db.models import Model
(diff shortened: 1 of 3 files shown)


intgr:
Closing this, looks like we'll be reversing the change in `django-stubs` instead:

* https://github.com/typeddjango/django-stubs/pull/1843

PR #509, opened 2023-11-15
Improve `BaseSerializer` and `ListSerializer` method `to_representation()`

intgr:
FYI @4c0n, this is a follow-up to #504. 

The original PR changed `BaseSerializer.to_representation()` to return `dict[str, Any]`.

However, turns out `BaseSerializer` subclass `ListSerializer` method actualy returns `list`.

Changed now:
* `BaseSerializer.to_representation()` -> `Any`
* `Serializer.to_representation()` -> `dict[str, Any]`
* `ListSerializer.to_representation()` -> `list[Any]`
* Also improved `ListSerializer.to_representation()` argument type

## Related issues

* #504


4c0n (approved):
Yep that seems right, I totally overlooked that part!

PR #511, opened 2023-11-17
Allow none as field value

### LINKED ISSUE #510: Field.to_representation() should accept None
# Bug report

## What's wrong
The type stubs do not allow `to_representation` to be given `None`, even though this is valid.

```python
from rest_framework import serializers
serializers.DateField(allow_null=True).to_representation(None)
```

mypy error:
```
Argument 1 to "to_representation" of "Field" has incompatible type "None"; expected "date"  [arg-type]
```

## How is that should be
`None` should be a supported value type for serializers.

## System information

- OS: PopOS 22.04
- `python` version: 3.11.4
- `django` version: 4.2.7
- `mypy` version: 1.6.1
- `django-stubs` version: 4.2.6

Kangaroux:
# I have made things!
This updates the field stubs to include `None` as valid input/output. The only projects that should be impacted by this change are those using serializer fields directly which I imagine are a very small minority. `BaseSerializer` is a subclass of `Field` but uses `Any` as the input/output types, so serializer stubs should remain unchanged.

## Related issues
- Closes #510

### DIFF
--- a/rest_framework-stubs/fields.pyi
+++ b/rest_framework-stubs/fields.pyi
@@ -430,7 +430,7 @@ class TimeField(Field[datetime.time | None, datetime.time | str | None, str | No
         allow_null: bool = ...,
     ) -> None: ...
 
-class DurationField(Field[datetime.timedelta| None, datetime.timedelta | str| None, str| None, Any]):
+class DurationField(Field[datetime.timedelta | None, datetime.timedelta | str | None, str | None, Any]):
     max_value: datetime.timedelta | None
     min_value: datetime.timedelta | None
     def __init__(
@@ -452,7 +452,7 @@ class DurationField(Field[datetime.timedelta| None, datetime.timedelta | str| No
         allow_null: bool = ...,
     ) -> None: ...
 
-class ChoiceField(Field[str| None, str | int | tuple[str | int, str | int | tuple]| None, str| None, Any]):
+class ChoiceField(Field[str | None, str | int | tuple[str | int, str | int | tuple] | None, str | None, Any]):
     html_cutoff: int | None
     html_cutoff_text: StrOrPromise | None
     allow_blank: bool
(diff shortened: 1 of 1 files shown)


rest_framework-stubs/fields.pyi
```diff
@@ -127,30 +127,20 @@ class Field(Generic[_VT, _DT, _RP, _IN]):
...
+        bool | None,
         Any,
     ]
 ):
     TRUE_VALUES: set[str | bool | int]
     FALSE_VALUES: set[str | bool | int | float]
     NULL_VALUES: set[str | None]
 
-class CharField(Field[str, str, str, Any]):
+class NullBooleanField(BooleanField): ...
```
Kangaroux, line 141: It looks like `NullBooleanField` was [deprecated in v3.14 of DRF](https://github.com/encode/django-rest-framework/blob/1db19f4b2d1500894448634852e02f73043382e4/docs/community/3.14-announcement.md#deprecations)

Do you want me to remove it in this PR or make a separate one?

intgr: Separate PR please.

rest_framework-stubs/fields.pyi
```diff
@@ -227,7 +217,7 @@ class SlugField(CharField):
 
 class URLField(CharField): ...
 
-class UUIDField(Field[uuid.UUID, uuid.UUID | str | int, str, Any]):
+class UUIDField(Field[uuid.UUID | None, uuid.UUID | str | int | None, str | None, Any]):
```
intgr, line 220: I don't think this is the best way to go about this.

I suspect not all methods using these typevars _can_ deal with `None`.

And many places that use these TypeVars already explicitly specify e.g. `_VT | None`. Seems a lot safer to just update the signatures of the `to_representation()` method.

intgr: Hrm, `IntegerField.to_representation()` actually does not allow `None`... Not sure what to make of this.

```python
    def to_representation(self, value):
        return int(value)
```

intgr (changes requested):

PR #518, opened 2023-12-08
Tests: Fix test after django-stubs StrPromise changes

intgr:
# I have made thangs!

Minor change in test error message output.

## Related issues

* Caused by https://github.com/typeddjango/django-stubs/pull/1841


intgr:
Merging myself as trivial change.

PR #520, opened 2023-12-08
Fix class hierachy and TypeVars of Serializer classes

intgr:
# I have made thunks!

* `ModelSerializer` was incorrectly deriving from `Serializer[Any]` and `BaseSerializer[T]`; replaced with `Serializer[T]`.
  This brings it into sync with DRF upstream.
* Added TypeVar to `HyperlinkedModelSerializer` class.
* Removed unnecessary `Generic[]` where implicit TypeVars are sufficient.

  https://mypy.readthedocs.io/en/stable/generics.html#defining-subclasses-of-generic-classes

  > If there are no Generic[...] in bases, then all type variables are collected in the lexicographic order (i.e. by first appearance).

### DIFF
--- a/rest_framework-stubs/relations.pyi
+++ b/rest_framework-stubs/relations.pyi
@@ -1,5 +1,5 @@
 from collections.abc import Callable, Iterable, Mapping, Sequence
-from typing import Any, Generic, TypeVar
+from typing import Any, TypeVar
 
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
Thank you!

intgr:
Oops, I missed the notification that this got apprpoved. Thanks

PR #530, opened 2024-01-04
Make `HyperlinkedRelatedField(view_name=)` & `HyperlinkedIdentityField(view_name=)` param optional

### LINKED ISSUE #537: Typing for classes inheriting HyperlinkedIdentityField
It seems that the following changes introduced a bug:
https://github.com/typeddjango/djangorestframework-stubs/commit/133e572f08c83c5048a88891686d399e496812f0

It’s possible to not pass any `view_name` to a `HyperlinkedRelatedField` if you create a subclass that will define the `view_name` directly.

For instance:
```python
class BookField(serializers.HyperlinkedRelatedField):
    view_name = 'book-detail'

    def get_queryset(self):
        return Book.objects.all()
```

is valid.
Right now mypy wrongfully complains with:
```
error: Missing positional argument "view_name" in call to "BookField"  [call-arg]
```

Alexerson:
Fixes #537 537


intgr (changes requested):
Thanks! Some allowlist entries can be removed from `scripts/stubtest/allowlist.txt`

See failed job output in https://github.com/typeddjango/djangorestframework-stubs/actions/runs/8345359816/job/22840080240?pr=530

intgr (approved):
Thanks! Sorry for the late review. :)

PR #541, opened 2024-02-13
Fixed `ModelSerializer.build_standard_field(model_field=)` parameter typing

noamkush:
This fixes the annotation for `ModelSerializer.build_standard_field` which takes the field instance and not the type.


intgr (approved):
Thanks, you are correct.

PR #546, opened 2024-02-26
allow floats for _GetDataType

asottile-sentry:
note that `float` in the mypy typing system includes `int` so including `| int` would be redundant


asottile-sentry:
I believe the test failures are unrelated and https://github.com/typeddjango/djangorestframework-stubs/pull/526 is needed to unbreak them

sobolevn (approved):
Thanks!

PR #596, opened 2024-04-23
Fix CI: suppress mypy errors in test suite that come from 3rd party packages

intgr:
`djangorestframework-stubs` CI was failing due to changes in `django-stubs` upstream.

This issue has bit us a few times in the past too. A suggestion was proposed in https://github.com/typeddjango/pytest-mypy-plugins/issues/134#issuecomment-1817723015 that should fix this once and for all, and we can remove hacks that I previously added in `mypy.ini`.


sobolevn (approved):
Thank you!

PR #600, opened 2024-04-25
Export imported modules in serializers.pyi. Fixes: #599

### LINKED ISSUE #599: Export imported modules in serializers.pyi
# Bug report

## What's wrong

When using validators in serializers, mypy raises an error

```py
class UserSerializer(serializers.ModelSerializer):
    class Meta:
        model = User
        validators = [
            serializers.UniqueTogetherValidator(
                queryset=User.objects.all(),
                fields=["username"],
            ),
            serializers.UniqueForDateValidator(
                queryset=User.objects.all(),
                field="email",
                date_field="date_joined",
            ),
        ]
```

```
error: Module "rest_framework.serializers" does not explicitly export attribute "UniqueTogetherValidator"  [attr-defined]
error: Module has no attribute "UniqueForDateValidator"  [attr-defined]
```

## How is that should be

According to [PEP-484 Stub Files](https://peps.python.org/pep-0484/#stub-files)

> Modules and variables imported into the stub are not considered exported from the stub unless the import uses the `import ... as ...` form or the equivalent `from ... import ... as ...` form.

The modules needs to be explicitly exported and the exports should be aligned with [rest_framework/serielizers](https://github.com/encode/django-rest-framework/blob/a677b09/rest_framework/serializers.py#L42)

```py
from rest_framework.validators import UniqueTogetherValidator as UniqueTogetherValidator
```

I'll happily open a PR in a bit

## System information

- OS: Kali GNU/Linux Rolling x86_64 6.6.9-amd64
- `python` version: 3.11
- `django` version: 4.2.7
- `mypy` version: 1.10.0
- `django-stubs` version: 4.2.7

SukiCZ:
# I have made things!

According to [PEP-484 Stub Files](https://peps.python.org/pep-0484/#stub-files) imported modules are not exported by default unless `from ... import ... as ...` is used. This exports `UniqueTogetherValidator` and aligns imports in `serializers`.

## Related issues

- Closes #599

Other imports should be exported as well. Following command prints 128 lines.

```bash
grep -r 'from rest_framework' rest_framework-stubs | grep -v ' as '
```

I know, some of those imports are not meant to be exported (e.g. they are not imported in `rest_framework` itself). I'm thinking about finding them automatically and maybe extend this PR for mode imports?

### DIFF
--- a/rest_framework-stubs/serializers.pyi
+++ b/rest_framework-stubs/serializers.pyi
@@ -62,8 +62,17 @@ from rest_framework.relations import RelatedField as RelatedField
 from rest_framework.relations import SlugRelatedField as SlugRelatedField
 from rest_framework.relations import StringRelatedField as StringRelatedField
 from rest_framework.utils.model_meta import FieldInfo, RelationInfo
-from rest_framework.utils.serializer_helpers import BindingDict, BoundField, ReturnDict, ReturnList
-from rest_framework.validators import BaseUniqueForValidator, UniqueTogetherValidator, Validator
+from rest_framework.utils.serializer_helpers import BindingDict as BindingDict
+from rest_framework.utils.serializer_helpers import BoundField as BoundField
+from rest_framework.utils.serializer_helpers import JSONBoundField as JSONBoundField
+from rest_framework.utils.serializer_helpers import NestedBoundField as NestedBoundField
+from rest_framework.utils.serializer_helpers import ReturnDict as ReturnDict
+from rest_framework.utils.serializer_helpers import ReturnList as ReturnList
+from rest_framework.validators import BaseUniqueForValidator, Validator
+from rest_framework.validators import UniqueForDateValidator as UniqueForDateValidator
+from rest_framework.validators import UniqueForMonthValidator as UniqueForMonthValidator
+from rest_framework.validators import UniqueForYearValidator as UniqueForYearValidator
+from rest_framework.validators import UniqueTogetherValidator as UniqueTogetherValidator
 from typing_extensions import Self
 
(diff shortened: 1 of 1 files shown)


rest_framework-stubs/serializers.pyi
```diff
@@ -62,8 +62,17 @@ from rest_framework.relations import RelatedField as RelatedField
...
+from rest_framework.utils.serializer_helpers import BoundField as BoundField
+from rest_framework.utils.serializer_helpers import JSONBoundField as JSONBoundField
+from rest_framework.utils.serializer_helpers import NestedBoundField as NestedBoundField
+from rest_framework.utils.serializer_helpers import ReturnDict as ReturnDict
+from rest_framework.utils.serializer_helpers import ReturnList as ReturnList
+from rest_framework.validators import BaseUniqueForValidator, Validator
+from rest_framework.validators import UniqueForDateValidator as UniqueForDateValidator
+from rest_framework.validators import UniqueForMonthValidator as UniqueForMonthValidator
+from rest_framework.validators import UniqueForYearValidator as UniqueForYearValidator
+from rest_framework.validators import UniqueTogetherValidator as UniqueTogetherValidator
```
sobolevn, line 75: Why do we need to export validators from serializers? Is it documented somewhere?

SukiCZ: I thought `serializers.UniqueTogetherValidator` is a valid syntax, similar to `serializers.CharField`. But it only applies to `fields` and `relations`, imported under [this line](https://github.com/encode/django-rest-framework/blob/a677b09/rest_framework/serializers.py#L47) :facepalm:

PR #601, opened 2024-04-26
Allow `None` for `ListField.min/max_length`

Viicos:
At runtime, this defaults to `None`:

```python
self.max_length = kwargs.pop('max_length', None)
self.min_length = kwargs.pop('min_length', None)
```


sobolevn (approved):
Thanks!

PR #610, opened 2024-05-10
Improve `DefaultRouter` constructor and attributes

intgr:
Noticed this while working on drf-nested-routers (https://github.com/alanjds/drf-nested-routers/pull/345)


sobolevn (approved):
Thanks!

PR #618, opened 2024-05-27
Accept Promise as Validator.message|missing_message

q0w:
# I have made things!
https://github.com/encode/django-rest-framework/blob/master/rest_framework/validators.py

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,7 @@ extend-exclude = [
 line-length = 120
 target-version = "py38"
 # See Rules in Ruff documentation: https://docs.astral.sh/ruff/rules/
+[tool.ruff.lint]
 select = [
     "B",        # bugbear
     "E",        # pycodestyle
@@ -30,7 +31,7 @@ select = [
 ]
 ignore = ["PYI021", "PYI024", "PYI041", "PYI043"]
 
-[tool.ruff.per-file-ignores]
+[tool.ruff.lint.per-file-ignores]
 "*.pyi" = [
     "B",
     "E501",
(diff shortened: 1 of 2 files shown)


sobolevn (approved):

rest_framework-stubs/validators.pyi
```diff
@@ -2,6 +2,7 @@ from collections.abc import Callable, Container, Iterable, MutableMapping
 from typing import Any, Protocol, TypeVar
 
 from django.db.models import Model, QuerySet
+from django.utils.functional import _StrOrPromise
```
intgr, line 5: Use `from django_stubs_ext import StrOrPromise`, which is public API.

pyproject.toml
```diff
@@ -14,6 +14,7 @@ extend-exclude = [
 line-length = 120
 target-version = "py38"
 # See Rules in Ruff documentation: https://docs.astral.sh/ruff/rules/
+[tool.ruff.lint]
```
intgr, line 17: Would you mind opening a separate MR about this?

intgr (changes requested):

rest_framework-stubs/validators.pyi
q0w: I checked django-stubs for that

pyproject.toml
q0w: Do you want to revert these changes?

intgr: No, I was seconds late with my review. No reason to revert if it's merged already.

rest_framework-stubs/validators.pyi
intgr: Yeah, that's internal to `django-stubs`. Other projects should import from `django_stubs_ext`

q0w: https://github.com/typeddjango/djangorestframework-stubs/pull/621

PR #621, opened 2024-05-28
Cleanup: Import StrOrPromise from django_stubs_ext

q0w:
# I have made things!
Maybe django_stubs_ext should be explicitly added here as a dependency?

## Related issues
https://github.com/typeddjango/djangorestframework-stubs/pull/618#discussion_r1616690441


intgr (approved):
Thanks for following up!

> Maybe django_stubs_ext should be explicitly added here as a dependency?

I discussed this with sobolevn in a PR some time ago and thought it's unnecessary; we can rely on the transitive dependency via `django-stubs`.

PR #632, opened 2024-07-02
Fix `SlugRelatedField.to_representation()` parameter name

DongwookKim0823:
# I have made things!

Fixes an inconsistency in `SlugRelatedField.to_representation()` parameter name from `value` to `obj`.

## Related issues

Upstream PR
- https://github.com/encode/django-rest-framework/pull/1800


intgr (approved):
Thanks!

PR #634, opened 2024-07-03
Fix `APIRequestFactory` method signatures

DongwookKim0823:
# I have made things!

I have updated the method signatures in `APIRequestFactory` to include the correct types and removed unnecessary variables.

- Removed the `follow: bool = ...` parameter from all methods.
- Updated the `options` method to accept `data: dict[str, str] | str | None`.


intgr (approved):
Thanks again!

PR #644, opened 2024-07-26
Add `objects` attribute to `Token`

sobolevn:
Yes, I remember now that adding explicit `objects` helps with `pyright`

### DIFF
--- a/rest_framework-stubs/authtoken/models.pyi
+++ b/rest_framework-stubs/authtoken/models.pyi
@@ -1,12 +1,14 @@
-from typing import Any
+from typing import Any, ClassVar
 
 from django.db import models
+from django.db.models.manager import Manager
 from typing_extensions import Self
 
 class Token(models.Model):
     key: models.CharField
     user: models.OneToOneField
     created: models.DateTimeField
+    objects: ClassVar[Manager[Self]]
     @classmethod
     def generate_key(cls) -> str: ...
 


sobolevn:
@flaeppe one more question. I really hope that this change won't affect user-defined models. (I think we would have lots of tests failing then, but still. Since many models don't have explicit managers defined).

flaeppe:
I'm not sure, was thinking about bisecting from #643 and see what commit I run into

flaeppe:
Oh, right, it was green 11 hours ago in https://github.com/typeddjango/djangorestframework-stubs/actions/runs/10099849613

flaeppe:
I don't have a repro other than this, but I think I figured it out:

The thing is that since https://github.com/typeddjango/django-stubs/pull/2276 now goes through the mro it conflicts with the 
 `mypy_django_plugin.transformers.models.MetaclassAdjustments` that removes `.objects` from `models.Model`. This in itself isn't a problem as long as we (or mypy) can enforce an ordering such that we always process `models.Model` before anything else. This nearly happens, it just seems to be that the `TypeInfo´ for `models.Model` isn't fully populated at that point. Then the lookup via the mro finds `objects` on `models.Model`, but a while afterwards `objects` is removed via `MetaclassAdjustments` and we get the error.

flaeppe:
`django-stubs` needs to provide a fix for this. This PR shouldn't be necessary when running the plugin(though it is for other cases).

I'm leaning towards reverting the plugin code that removes `Model.objects`. Most users seem to run with default behaviour of having an `objects` manager, we have multiple issues with different angles in `django-stubs` regarding this. And to follow Django's behaviour for other type checkers we'd probably want to declare e.g. `objects` as an _abstract attribute_. But there's currently no support for abstract attributes in Python

flaeppe:
@sobolevn I've opened https://github.com/typeddjango/django-stubs/pull/2280

sobolevn:
This PR is still fine, because it will also help `pyright` users, which don't use our plugin.

PR #649, opened 2024-08-10
Add accepted_renderer, accepted_media_type to Request and Response

q0w:
# I have made things!
ref: https://github.com/encode/django-rest-framework/blob/f113ab6b68e5ab8c395a93f168fe46c77c05b324/rest_framework/views.py#L434-L435
ref: https://github.com/encode/django-rest-framework/blob/f113ab6b68e5ab8c395a93f168fe46c77c05b324/rest_framework/request.py#L115-L118


rest_framework-stubs/request.pyi
```diff
@@ -47,6 +48,8 @@ class Request(HttpRequest):
     parsers: Sequence[BaseParser] | None
     authenticators: Sequence[BaseAuthentication | ForcedAuthentication] | None
     negotiator: BaseContentNegotiation | None
+    accepted_renderer: BaseRenderer | None
+    accepted_media_type: str | None
```
intgr, line 52: I don't think these can actually be `None`? Looking at `APIView.perform_content_negotiation()`, it rather raises exception than returns `None`.

Same for `parse_header_parameters()`, it raises `NotAcceptable` if it can't find a match.

But still this PR is an improvement, I'll merge this and get a release out the door. We can improve this later.

q0w: but request may not have this attrs at all. There are hasattr checks after all

intgr: Yeah, there's no way to annotate 'possibly missing attribute' for classes. A code comment in pyi is best we can do.

But that's different from None. If the value may be missing but never None then I'd suggest to remove `| None`. Because that's even more misleading.

PR #651, opened 2024-08-13
Set configured user model as `request.user`

### LINKED ISSUE #650: New mypy failures on `djangorestframework-stubs` after `django-stubs` update
```

Run mypy --cache-dir=/dev/null --no-incremental rest_framework-stubs
rest_framework-stubs/authtoken/models.pyi:16: error: Name "pk" already defined (possibly by an import)
rest_framework-stubs/authtoken/models.pyi:17: error: Cannot override writeable attribute with read-only property
rest_framework-stubs/request.pyi:71: error: Signature of "user" incompatible with supertype "HttpRequest"
rest_framework-stubs/request.pyi:71: note:      Superclass:
rest_framework-stubs/request.pyi:71: note:          Union[User, AnonymousUser]
rest_framework-stubs/request.pyi:71: note:      Subclass:
rest_framework-stubs/request.pyi:71: note:          Union[AbstractBaseUser, AnonymousUser]
Found 3 errors in 2 files (checked 68 source files)
```

cc @flaeppe 

Looks like that it fails because of our recent User plugin change.

flaeppe:
Uses the placeholder `_UserModel` from `django-stubs` to hook in to its plugin to get the configured model.

## Related issues

- Closes #650

### DIFF
--- a/rest_framework-stubs/request.pyi
+++ b/rest_framework-stubs/request.pyi
@@ -4,7 +4,7 @@ from contextlib import AbstractContextManager, contextmanager
 from types import TracebackType
 from typing import Any
 
-from django.contrib.auth.base_user import AbstractBaseUser
+from django.contrib.auth.base_user import AbstractBaseUser, _UserModel
 from django.contrib.auth.models import AnonymousUser
 from django.http import HttpRequest
 from django.http.request import _ImmutableQueryDict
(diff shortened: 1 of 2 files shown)


flaeppe:
This fixes the supertype issue, but I'm not sure what to make of the other one with the proxy model. I suppose a `type: ignore[...]` might be reasonable there since it actually is a property in DRF

sobolevn (approved):
Thank you! I think that `UserModel` should be in `django-stub-ext`. It should be public.

flaeppe:
Yes indeed. I think it should be public too

PR #664, opened 2024-09-13
Add Stack Overflow badge as well

sobolevn:
I think that it is appropriate for this project as well.


README.md
```diff
@@ -5,6 +5,7 @@
 [![Build status](https://github.com/typeddjango/djangorestframework-stubs/workflows/test/badge.svg?branch=master&event=push)](https://github.com/typeddjango/djangorestframework-stubs/actions?query=workflow%3Atest)
 [![Checked with mypy](https://www.mypy-lang.org/static/mypy_badge.svg)](https://mypy-lang.org/)
 [![Gitter](https://badges.gitter.im/mypy-django/Lobby.svg)](https://gitter.im/mypy-django/Lobby)
+[![StackOverflow](https://shields.io/badge/ask-stackoverflow-orange?logo=stackoverflow)](https://stackoverflow.com/questions/tagged/django-stubs?tab=Active)
```
intgr, line 8: I initially thought it should use a SO tag different from `django-stubs`. But then again, they are quite related; and there are no questions for the `djangorestframework-stubs` tag: https://stackoverflow.com/questions/tagged/djangorestframework-stubs.

I don't mind using the same tag.

intgr (approved):

PR #672, opened 2024-09-27
Improve `GenericAPIView.serializer_class` classvar type generics

### LINKED ISSUE #673: GenericAPIView serializer_class type should use _MT_co
# Bug report

## What's wrong

The typing is a little incorrect as the truthy type does not match what the `get_serializer_class` returns.
## How is that should be

 `serializer_class` probably should be: `type[BaseSerializer[_MT_co]] | None` so that the truthy object type matches what `get_serializer_class` returns.

## System information

- OS:
- `python` version:
- `django` version:
- `mypy` version:
- `django-stubs` version:

jheld:
# I have made things!

The type for the attribute does not match what the `get_serializer_class` function returns.

Thus, modifying it to match so that the truthy type of it is `type[BaseSerializer[_MT_co]]`

Unfortunately this doesn't solve the issue of projects programmatically returning different serializers in `get_serializer_class` such that one option would be: `return self.serializer_class` such that it would be technically incompatible with the typing. But that seems outside the scope of this change and there are other ways to accomplish it (e.g. holding known serializer classes in a `dict` on the class as an attribute).

## Related issues

- Closes #673

### DIFF
--- a/rest_framework-stubs/generics.pyi
+++ b/rest_framework-stubs/generics.pyi
@@ -33,7 +33,7 @@ class BaseFilterProtocol(Protocol[_MT_inv]):
 
 class GenericAPIView(views.APIView, UsesQuerySet[_MT_co]):
     queryset: QuerySet[_MT_co] | Manager[_MT_co] | None
-    serializer_class: type[BaseSerializer] | None
+    serializer_class: type[BaseSerializer[_MT_co]] | None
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
I think that the change is correct.

intgr (approved):
Looks good :+1:

PR #686, opened 2024-10-28
represent choices in ChoiceField as Sequence

### LINKED ISSUE #454: consider `Sequence[Any]` instead of `Iterable[Any]` for `ChoiceField`s `choices`
this isn't a bug but perhaps a suggestion.

when rest_framework's `ChoiceField` and `MultipleChoiceField` receive an unordered collection for `choices` it can lead to a negative experience: unordered form output, unordered schema, etc.

https://github.com/typeddjango/djangorestframework-stubs/blob/e0e9e1f6ba9411345d00abb904b62c106ca8d7b1/rest_framework-stubs/fields.pyi#L460

we've made a small plugin for `sentry` that enforces this for our uses since we kept hitting this problem:  https://github.com/getsentry/sentry/pull/53707

though this isn't entirely in spirit of the type system so understandable if this isn't a good idea

asottile:
this enforces that choices are in a deterministic order for forms / schemas / etc.

# I have made things!

## Related issues

resolves #454

### DIFF
--- a/rest_framework-stubs/fields.pyi
+++ b/rest_framework-stubs/fields.pyi
@@ -463,7 +463,7 @@ class ChoiceField(Field[str, str | int | tuple[str | int, str | int | tuple], st
     _choices: dict
     def __init__(
         self,
-        choices: Iterable[Any],
+        choices: Sequence[Any],
         *,
         read_only: bool = ...,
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thank you!

PR #704, opened 2024-12-04
Update rest_framework.schemas.openapi.ExternalDocumentationObject

kwang1215:
# I have made things!

### DIFF
--- a/rest_framework-stubs/schemas/openapi.pyi
+++ b/rest_framework-stubs/schemas/openapi.pyi
@@ -18,10 +18,6 @@ class DRFOpenAPIInfo(TypedDict, total=False):
     version: str
     description: str
 
-class ExternalDocumentationObject(TypedDict, total=False):
-    description: str
-    url: str
-
 class DRFOpenAPISchema(TypedDict, total=False):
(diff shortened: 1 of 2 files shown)


sobolevn (commented):
I don't think that this change is correct. Can you please provide a link to the source code that proves that `externalDocs` key was removed.

kwang1215:
> I don't think that this change is correct. Can you please provide a link to the source code that proves that `externalDocs` key was removed.

I couldn't find the ExternalDocumentationObject class and externalDocs in the rest_framework/schemas/openapi.py file. However, after looking at the commit where ExternalDocumentationObject was first added, I found information related to externalDocs on https://github.com/OAI/OpenAPI-Specification.git.

Is there a connection between rest_framework/schemas/openapi.py and https://github.com/OAI/OpenAPI-Specification.git? 
I am still a beginner developer, so I'm not sure.

sobolevn:
Yes, this is a `@type_check_only` `TypedDict` object. It represents the OpenAPI JSON spec.

sobolevn:
I think that it is safe to close this now :)

PR #705, opened 2024-12-04
Update rest_framework.renderers.BaseRenderer.format, media_type

Limsunoh:
# I have made things!


rest_framework-stubs/renderers.pyi
```diff
@@ -13,8 +13,8 @@ from rest_framework.views import APIView
 def zero_as_none(value: Any) -> Any | None: ...
 
 class BaseRenderer:
-    media_type: str
-    format: str
+    media_type: str | None
```
sobolevn, line 16: If you are to make this change - you should also change all its subclasses: https://github.com/encode/django-rest-framework/blob/dbac145638758413b966c3418fa5f3f651e3e02a/rest_framework/renderers.py#L42-L48

They have to override `None` to some `str` value.

sobolevn (changes requested):

rest_framework-stubs/renderers.pyi
Limsunoh: Thanks, I must have missed something. 
Thanks for checking it out.

sobolevn:
(misclick)

intgr (commented):
Sorry, I'm a little apprehensive about this change.

I think I had a look at this some time ago and came to the conclusion that any **actual** subclass of `BaseRenderer` must assign strictly non-`None` value to the `media_type` field. If it's None, DRF would actually crash.

So `None` is actually not a valid value to these fields. It's just a placeholder in the "abstract base class".

Typehinting this as `str | None` might mislead someone who tries to implement their own `Renderer`.

🟢 But what we should do is add comments explaining this & move the `allowlist_todo.txt` entry into `allowlist.txt`.

intgr:
Documented this discrepancy in #833.

Issue #718, opened 2025-01-13
Incorrect Type Hint for ModelSerializer class `instance` attribute


JoshFerge:
`instance` is not a `Sequence` when used in ModelSerializer code, I believe this type hint was in error. https://github.com/typeddjango/djangorestframework-stubs/blob/c78da6d315f0994ad472d2c2406ae5718ef36057/rest_framework-stubs/serializers.pyi#L195
it should just be `instance: _MT | None`

JoshFerge:
see https://github.com/typeddjango/djangorestframework-stubs/pull/719#issuecomment-2588394634

PR #719, opened 2025-01-13
[Breaking change] Remove work-around for `ModelSerializer.instance` field for `many=True`

JoshFerge:
fixes https://github.com/typeddjango/djangorestframework-stubs/issues/718

### Edit by @intgr

The background of this change is described in #827. If this change caused a regression in your use case, please comment there ➔ #827

### DIFF
--- a/rest_framework-stubs/serializers.pyi
+++ b/rest_framework-stubs/serializers.pyi
@@ -192,7 +192,7 @@ class ModelSerializer(Serializer[_MT]):
     serializer_url_field: type[RelatedField]
     serializer_choice_field: type[Field]
     url_field_name: str | None
-    instance: _MT | None  # type: ignore[assignment]
+    instance: _MT | None
 
     class Meta:
         model: type[_MT]  # type: ignore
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
I could not find any `Sequence` usages in https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py

Thanks!

JoshFerge:
> I could not find any `Sequence` usages in https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py
> 
> Thanks!

just fixed the type ignore error in ci

intgr:
`instance` is a sequence when the serializer is initialized with `many=True`. Although it truly is annoying to deal with.

There was a plan to fix this with an overloaded Serializer __new__ method, but it requires changes to mypy:

* https://github.com/python/mypy/issues/15182

JoshFerge:
got it. thanks for clarifying! will close.

JoshFerge:
Re-opening, as I think that _internally_ the type will never be a sequence, as the `new function will return a `ListSerializer` if a sequence is passed in. I think this is correct.

As written, the serializer is impossible to write a subclass for.

intgr:
Sorry for the lack of reply. It's a tricky decision between two bad options and I've been pondering this in the background. I'm rather inclined to merge this.

asottile-sentry:
actually a simpler patch here is to delete the line that's edited (since BaseSerializer already sets `instance: _IN | None`)

PR #756, opened 2025-04-26
Restrict setuptools version to <79.0.0

davidmannion:
setuptools v79.0.0 removed support for `legacy-editable` installs, causing CI to fail.

This was fixed in django-stubs by restricting the setuptools version to <79.0.0 (see https://github.com/typeddjango/django-stubs/pull/2617, thanks @michalpokusa). This PR applies the same fix.


sobolevn (approved):

sobolevn:
The CI still fails, please, feel free to fix it in another PR :)

PR #758, opened 2025-04-27
Fix missing TypeAlias for `request.user`

### LINKED ISSUE #757: `request.user` is inferred as `Any | AnonymousUser` after upgrading django-stubs to 5.2.0
# Bug report

## What's wrong

`request.user` is inferred as `Any | AnonymousUser` after upgrading django-stubs to 5.2.0. 

https://github.com/typeddjango/djangorestframework-stubs/pull/651 used `_UserModel` from django-stubs, allowing mypy to use the correct type for custom User models. https://github.com/typeddjango/django-stubs/commit/8249ccd92776ed0917be59e3923100eea2c6187a refactored the User `TypeVars`, removing `_UserModel`.

## How is that should be

`request.user` should be `User | AnonymousUser`.

## System information

- OS: Ubuntu 24.04
- `python` version: 3.13
- `django` version: 5.2
- `mypy` version: 1.15.0
- `django-stubs` version: 5.2.0

I've fixed this locally and I'm happy to submit a PR, but I thought it might be a bit too presumptuous to also bump the django-stubs version (and mypy version to match that used in django-stubs)!

davidmannion:
`_UserModel` was replaced with `_User` in `django-stubs` 5.2.0.

## Related issues

- Closes #757

### DIFF
--- a/rest_framework-stubs/request.pyi
+++ b/rest_framework-stubs/request.pyi
@@ -4,8 +4,8 @@ from contextlib import AbstractContextManager, contextmanager
 from types import TracebackType
 from typing import Any
 
-from django.contrib.auth.base_user import AbstractBaseUser, _UserModel
-from django.contrib.auth.models import AnonymousUser
+from django.contrib.auth.base_user import AbstractBaseUser
+from django.contrib.auth.models import AnonymousUser, _User
 from django.http import HttpRequest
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thanks! LGTM

PR #791, opened 2025-07-17
Tests: Fix compatbility with mypy 1.17 due to bump in django-stubs

intgr:
Mypy version was bumped in django-stubs:

* https://github.com/typeddjango/django-stubs/pull/2744

The `force_uppercase_builtins` setting was deprecated in mypy 1.17. Caused test failures due to new logging.

https://mypy.readthedocs.io/en/stable/changelog.html#deprecated-flag-force-uppercase-builtins

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -13,7 +13,6 @@ disallow_incomplete_defs = true
 disable_error_code = empty-body
 # TODO: update our test error messages to match new mypy output
 show_error_codes = false
-force_uppercase_builtins = true
 force_union_syntax = true
 
 plugins =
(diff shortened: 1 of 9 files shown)


mypy.ini
```diff
@@ -13,7 +13,6 @@ disallow_incomplete_defs = true
 disable_error_code = empty-body
 # TODO: update our test error messages to match new mypy output
 show_error_codes = false
-force_uppercase_builtins = true
 force_union_syntax = true
```
sobolevn, line 16:
```suggestion
```

let's remove it as well.

sobolevn (commented):
Thanks! This proposal is optional. If it hard to do, feel free to ignore.

mypy.ini
intgr: Agreed, I was planning to do it in a separate PR.

PR #792, opened 2025-07-17
Tests: Cleanup usages of obsolete Union/Optional types

intgr:
We can now disable `force_union_syntax` in mypy settings.

* Suggested in https://github.com/typeddjango/djangorestframework-stubs/pull/791
* Already fixed in django-stubs in https://github.com/typeddjango/django-stubs/pull/2744


sobolevn (approved):
Thank you!

PR #797, opened 2025-08-04
 Switch to `uv` to manage dev dependencies 

intgr:
## What was done

I and @CodingWithSaksham adapted `djangorestframework-stubs` to the `uv` package manager, like `django-stubs`.

* Using build backend `uv_build` -- as opposed to `hatchling` which is used by `django-stubs`
* `setup.py` and `setup.cfg` migrated into `pyproject.toml`
  * Removed obsolete flake8 config.
* `requirements.txt` merged into `pyproject.toml`
* `pytest.ini` merged into `pyproject.toml`
* GitHub `test.yml` workflows
  * Using `astral-sh/setup-uv@v6` to install uv & Python
  * Converted to use `uv run` & other `uv` commands.
  * New check to ensure `uv.lock` is up to date
* No longer depending on `setuptools` -- closes #761

## Related `django-stubs` PRs:

* https://github.com/typeddjango/django-stubs/pull/2688
* https://github.com/typeddjango/django-stubs/pull/2698


intgr:
@UnknownPlatypus would you be interested in looking into this?

UnknownPlatypus:
@intgr probably not until next week but I can have a look yes 👍

CodingWithSaksham:
I have made progress with this but the tests keep failing. I have tried everything that I could. Here is a sample of the error code

```bash
E   pytest_mypy_plugins.utils.TypecheckAssertionError: Output is not expected: 
E   Actual:
E     main:1: error: Skipping analyzing "rest_framework.permissions": module is installed, but missing library stubs or py.typed marker  [import-untyped] (diff)
E     main:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports (diff)
E     main:2: error: Skipping analyzing "rest_framework.decorators": module is installed, but missing library stubs or py.typed marker  [import-untyped] (diff)
E     main:9: error: Unsupported operand type for ~ ("type[Permission]")  [operator] (diff)
E   Expected:
E     (empty)
```

Every test is giving nearly the same error but the line 
```bash
E Actual:
E     main:1: error: Skipping analyzing "rest_framework.permissions": module is installed, but missing library stubs or py.typed marker  [import-untyped] (diff)
```
is repeated in every one, even though I have added a `py.typed` file in `rest_framework-stubs` directory.

I have configured `pyproject.toml` with this for `mypy` to use
```toml
[tool.mypy]
mypy_path = ["rest_framework-stubs"]
allow_redefinition = true
check_untyped_defs = true
incremental = true
strict_optional = false
show_traceback = true
warn_redundant_casts = true
warn_unused_configs = true
warn_unreachable = true
disallow_untyped_defs = true
namespace_packages = true
disable_error_code = "empty-body"
show_error_codes = false
plugins = ["mypy_django_plugin.main", "mypy_drf_plugin.main"]
```

Could you please guide me with this?

CodingWithSaksham:
Haven't pushed it yet! Will push it today for you to take a look at!

CodingWithSaksham:
Pushed the code! @intgr 
Here's the repo link!

https://github.com/CodingWithSaksham/djangorestframework-stubs/tree/use-uv

intgr:
@CodingWithSaksham Thanks! I took your branch and managed to make it build successfully with some changes.

I lost track of exactly who made what changes, is it OK if I credit you with a `Co-authored-by:` commit message tag?

* Switching from `hatchling` build system to `uv`'s own `uv_build` seemed to solve the main issue.
* Configuring mypy in `pyproject.toml` caused configuration to be ignored by `pytest_mypy_plugins` -- moved it back into `mypy.ini` for now. This migration can be done separately.
* Updated all dependency versions to match those in latest `master` branch.

CONTRIBUTING.md
```diff
@@ -39,13 +39,17 @@ In order to be able to continously sync your fork with the origin repository's m
...
+# TODO FIXME needs updating
+
 ```bash
 python3 -m venv .venv
 source .venv/bin/activate
 ```
 
 Then install the dev requirements:
 
+# TODO FIXME needs updating
```
intgr, line 51: Will update CONTRIBUTING instructions soon

CodingWithSaksham:
I am absolutely fine with this! Thanks for the help!

.github/workflows/test.yml
```diff
@@ -24,89 +27,78 @@ jobs:
...
       - name: Install dependencies
-        run: |
-          pip install -U pip "setuptools<79.0.0" wheel
-          SETUPTOOLS_ENABLE_FEATURES=legacy-editable pip install -r ./requirements.txt
-
+        run: uv sync
       - name: Run tests
-        # Suppress errors from other packages due to https://github.com/typeddjango/pytest-mypy-plugins/issues/134
-        run: pytest --mypy-only-local-stub
+        run: uv run pytest
```
intgr, line 66: `--mypy-only-local-stub` moved to `pyproject.toml` pytest config.

pyproject.toml
```diff
@@ -1,34 +1,116 @@
...
+testpaths = ["tests"]
+addopts = [
+  "--tb=native",
+  "-s",
+  "-v",
+  "--cache-clear",
+  "--mypy-extension-hook=scripts.tests_extension_hook.django_plugin_hook",
+  # Suppress errors from other packages due to https://github.com/typeddjango/pytest-mypy-plugins/issues/134
+  "--mypy-only-local-stub",
+  "--mypy-ini-file=mypy.ini"
```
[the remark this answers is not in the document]
intgr: This turned out to be problematic. Maybe later.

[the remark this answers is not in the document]
CodingWithSaksham: Did this previously, it made none of the tests pass. We need to approach this carefully then.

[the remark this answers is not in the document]
CodingWithSaksham: Updates my code to use `uv-build` instead of `hatch` and 29 out of the 36 test cases are failing!

Seems like `pytest` is unable to find 
```toml
disable_error_code = ["empty-body"]
enable_error_code = ["ignore-without-code"]
```

when written in `pyproject.toml`

pyproject.toml
```diff
@@ -1,34 +1,116 @@
...
+  "types-Pygments==2.19.0.20250809",
+  "types-pyyaml==6.0.12.20250915",
+  "django-stubs[compatible-mypy] @ git+https://github.com/typeddjango/django-stubs",
+  "django-stubs-ext @ git+https://github.com/typeddjango/django-stubs#subdirectory=ext",
+  "djangorestframework-stubs[compatible-mypy,coreapi,markdown]",
+]
+
+[build-system]
+requires = ["uv_build>=0.8.19,<0.9.0"]
+build-backend = "uv_build"
```
UnknownPlatypus, line 61: Can you check `python -m build` produces the same wheel and sdist compared to before ?

intgr: Good idea. I compared `dist` results when building `master` branch to the one produced by `uv build` in my branch. The differences look insignificant.

* There are frivolous differences in `PKG-INFO`, like different order of fields; package name casefolding; whether single quotes or double quotes are used, etc
* The `uv` version of tarball is missing the whole `djangorestframework_stubs.egg-info` directory
* The `uv` version of wheel is missing a top-level file `top_level.txt` which lists top-level directory names.

image

Diff of `djangorestframework_stubs-3.16.3-py3-none-any.whl`:

image

Diff of `djangorestframework_stubs-3.16.3.tar.gz`

image

I posted the full diff here in case you're curious: https://gist.github.com/intgr/0f25a31b859ffe4ce6f0fe157266c00b

intgr: Actually I'm not sure whether `egg-info` is significant. But it seems to largely duplicate `PKG-INFO`. Also what's `top_level.txt` used for?

CodingWithSaksham: `top_level.txt` is added by `setuptools` during its `egg-info` phase and it tells what top level modules/packages are installed.

Since `uv` is not making an `egg-info` as its not bounded by `setuptools`, I think that file is irrelevant.

intgr: Looks like `.egg-info` is [considered obsolete together with the whole egg package format](https://packaging.python.org/en/latest/discussions/package-formats/#what-about-eggs).

uv.lock
```diff
@@ -0,0 +1,1098 @@
...
+]
+sdist = { url = "https://files.pythonhosted.org/packages/4c/8c/2a21594337250a171d45dda926caa96309d5136becd1f48017247f9cdea0/django-5.2.6.tar.gz", hash = "sha256:da5e00372763193d73cecbf71084a3848458cecf4cee36b9a1e8d318d114a87b", size = 10858861, upload-time = "2025-09-03T13:04:03.23Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/f5/af/6593f6d21404e842007b40fdeb81e73c20b6649b82d020bb0801b270174c/django-5.2.6-py3-none-any.whl", hash = "sha256:60549579b1174a304b77e24a93d8d9fafe6b6c03ac16311f3e25918ea5a20058", size = 8303111, upload-time = "2025-09-03T13:03:47.808Z" },
+]
+
+[[package]]
+name = "django-stubs"
+version = "5.2.5"
+source = { git = "https://github.com/typeddjango/django-stubs#f0419144b6dcdf44198b3be5f6994194ab6a94a5" }
```
intgr, line 176: One thing I realized is that `uv.lock` now also freezes the `django-stubs` repo commit & it won't update automatically. So this PR as-is would prevent finding out when `django-stubs` changes have broken `djangofestframework-stubs`. That would be bad.

Options:

1. Add `uv lock --upgrade-package=django-stubs --upgrade-package=django-stubs-ext` to workflows, which would restore the status quo: if a breaking `django-stubs` change is made, it will break our workflows without clear indication that was the cause. That was never ideal, this would be simplest for now.
2. Maybe teach @dependabot to update this. That would cause a lot of PRs & noise though, every time there's a commit to `django-stubs`. I would prefer not.
3. Ideally we would test every `master` build with latest `django-stubs` & only open a PR when breakage is detected. But I'm not sure how to set that up.

CodingWithSaksham: For the 3rd one, we could make a `cron` job or have `django-stubs` make a release webhook

The GitHub action will take the new `django-stubs` and run our tests against it, if it fails we can open a PR

UnknownPlatypus: > Ideally we would test every master build with latest django-stubs & only open a PR when breakage is detected. But I'm not sure how to set that up.

We can install django-stubs@master in ci (I attempted that for mypy in [django-stubs ](https://github.com/typeddjango/django-stubs/pull/2671/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88)) or could setup a daily ci job ( see this ruff [daily fuzzing](https://github.com/astral-sh/ruff/blob/main/.github/workflows/daily_fuzz.yaml) job for ex)

intgr: @sobolevn I've requested to enable the [Renovate app](https://github.com/apps/renovate) in this repo.

UnknownPlatypus (approved):
This is good for me, thanks for the work on this!

sobolevn:
Done.

PR #825, opened 2025-09-16
Accept lazy strings as path in `APIRequestFactory` and `APIClient` http methods

browniebroke:
# I have made things!

I received and update for mypy 1.17.1 -> 1.18.1 and started to see some failures pointing at code like this in my tests:

```python
class TestAPIView:
    url = reverse_lazy("api:example")

    def test_ok(self, api_client: APIClient):
        ...
        response = api_client.get(self.url)  # Error here
        ...
```

The error message was:

```
error: Argument 1 to "get" of "APIClient" has incompatible type "_StrPromise"; expected "str"  [arg-type]
```

The code runs fine and the type check was ok with it in previous versions. Therefore we can probably safely accept lazy strings as path.

~Note: while running the tests it seems like we also need to update the base client class in django-stubs. I'll try to do that at some point when I have more time~ EDIT: it was due to a suppression comment being on the wrong place.

### DIFF
--- a/scripts/stubtest/allowlist_todo.txt
+++ b/scripts/stubtest/allowlist_todo.txt
@@ -14,7 +14,6 @@ rest_framework.authtoken.models.TokenProxy.get_next_by_created
 rest_framework.authtoken.models.TokenProxy.get_previous_by_created
 rest_framework.authtoken.models.TokenProxy.pk
 rest_framework.authtoken.models.TokenProxy.user_id
-rest_framework.authtoken.views.ObtainAuthToken.serializer_class
 rest_framework.authtoken.views.obtain_auth_token
 rest_framework.compat.CodeBlockPreprocessor
 rest_framework.compat.Preprocessor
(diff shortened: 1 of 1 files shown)


tests/typecheck/test_test.yml
```diff
@@ -25,3 +25,44 @@
...
+  - method: patch
+  - method: delete
+  - method: options
+  main: |
+    from django.urls import reverse_lazy
+    from rest_framework import test, status
+
+    url = reverse_lazy("api:example")
+    request_factory = test.APIRequestFactory()
+    request_factory.{{ method }}(url)
```
intgr, line 43: Optional nit: the test runner is quite slow to set up each test case, it's better to merge this a s a single test, e.g.

```suggestion
    request_factory.get(url)
    request_factory.post(url)
    request_factory.put(url)
    ...
```

browniebroke:
The stubtest failure seems unrelated... AFAIK it happens on other branches. Seems like an easy fix though

PR #852, opened 2025-10-14
Fix `Field.get_attribute()` return and `RelatedField.to_representation()` parameter for `PKOnlyObject`

### LINKED ISSUE #829: Wrong return type for Field.get_attribute()
# Bug report

## What's wrong

The return type of `Field.get_attribute()` is declared as `_RP`, but I think it should be `_VT`.

## How is that should be

Looking at the implementation of `Serializer.to_representation()`, the value returned by `Field.get_attribute()` is passed to `Field.to_representation()`, whose argument is declared to be `_VT`. Thus I think the declaration of `Field.get_attribute()` should be changed to:

```python
def get_attribute(self, instance: _IN) -> _VT | None: ...
```

## System information

- OS: macOS 14.8
- `python` version: Python 3.12.11
- `django` version: 4.2.24
- `mypy` version: 1.15.0
- `django-stubs` version: 5.2.5

emmanuel-ferdman:
## PR Summary
This PR adjusts the typing for relational serializer fields so they return the actual model instances or the lightweight `PKOnlyObject` Django REST Framework sometimes uses, instead of the already-serialized values. This matches DRF’s real behavior: `get_attribute` hands `to_representation` either the full object or the PK-only placeholder, so the new annotations describe exactly what callers receive and avoid type errors when the optimization kicks in.

Fixes #829.


rest_framework-stubs/relations.pyi
```diff
@@ -68,7 +68,7 @@ class RelatedField(Field[_MT, _DT, _PT, Any]):
     @property
     def grouped_choices(self) -> dict: ...
     def iter_options(self) -> Iterable[Option]: ...
-    def get_attribute(self, instance: _MT) -> _PT | None: ...
+    def get_attribute(self, instance: _MT) -> _MT | PKOnlyObject | None: ...
```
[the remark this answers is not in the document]
emmanuel-ferdman: @sobolevn added a few tests, thanks for the review 😄

rest_framework-stubs/relations.pyi
```diff
@@ -31,7 +31,7 @@ _MT = TypeVar("_MT", bound=Model)
 _DT = TypeVar("_DT")  # Data Type
 _PT = TypeVar("_PT")  # Primitive Type
 
-class RelatedField(Field[_MT, _DT, _PT, Any]):
+class RelatedField(Field[_MT | PKOnlyObject, _DT, _PT, Any]):
```
sobolevn:
```suggestion
class RelatedField(Field[_MT, _DT, _PT, Any]):
```

This does not seem correct to me.
CC @UnknownPlatypus

UnknownPlatypus: Sorry I have very little experience with `djangorestframework-stubs`, I'm not sure I'll be able to really help here

tests/typecheck/test_serializers.yml
```diff
@@ -49,6 +49,54 @@
...
+    obj = PKOnlyObject(pk=1)
+    reveal_type(field.to_representation(obj))  # N: Revealed type is "builtins.str"
+
+- case: related_field_validators_accept_model_type
+  main: |
+    from django.contrib.auth.models import User
+    from rest_framework import serializers
+
+    def user_validator(value: User) -> None:
+        _ = value.username
```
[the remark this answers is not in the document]
sobolevn: And the same for the second test

tests/typecheck/test_serializers.yml
```diff
@@ -49,6 +49,54 @@
...
+
+    field = serializers.PrimaryKeyRelatedField(
+        queryset=User.objects.all(),
+        validators=[user_validator]
+    )
+
+- case: related_field_default_is_model_type
```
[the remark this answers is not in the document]
emmanuel-ferdman: I added them previously to prove that we don't need to add `PKOnlyObject` to the base type parameter again. But we don't need them anymore.

rest_framework-stubs/relations.pyi
```diff
@@ -68,7 +68,7 @@ class RelatedField(Field[_MT, _DT, _PT, Any]):
     @property
     def grouped_choices(self) -> dict: ...
     def iter_options(self) -> Iterable[Option]: ...
-    def get_attribute(self, instance: _MT) -> _PT | None: ...
+    def get_attribute(self, instance: _MT) -> _MT | PKOnlyObject | None: ...  # type: ignore[override]
```
intgr: Since `use_pk_only_optimization() -> False` in `HyperlinkedRelatedField` and `StringRelatedField`, they don't return `PKOnlyObject`, should we override their `get_attribute()` back to return `_MT | None`?

emmanuel-ferdman: I'm thinking to skip this since I think `HyperlinkedRelatedField` can still return `PKOnlyObject` when `lookup_field='pk'` (making the narrowing incorrect for that case), and the precision gain for `StringRelatedField` alone seems small compared to the maintenance cost of the extra override - but happy to try and implement if you think it's worth it.

intgr (approved):

PR #856, opened 2025-10-21
CI: Add pre-commit config to update `uv.lock`

intgr:
* Also `django-stubs`: https://github.com/typeddjango/django-stubs/pull/2881


intgr:
Looks like it doesn't work because there's no network access from `pre-commit.ci`? Not sure whether it's a bug or on purpose.

PR #946, opened 2026-04-02
Change `Serializer` context type `dict` -> `Mapping`

### LINKED ISSUE #636: Allow mapping as serializer context
`TypedDict` t is not compatible with `dict`, see https://peps.python.org/pep-0589/#type-consistency. However, they are very practical, for instance, to type hint the context of serializers that expect more arguments.

Using a `Mapping` for the type would give more flexibility.

BrianHicks:
# I have made things!

This allows passing in `TypedDict` as well as ordinary dictionaries.

## Related issues

Closes #636


rest_framework-stubs/generics.pyi
```diff
@@ -40,7 +40,7 @@ class GenericAPIView(views.APIView, UsesQuerySet[_MT_co]):
     def get_object(self) -> _MT_co: ...
     def get_serializer(self, *args: Any, **kwargs: Any) -> BaseSerializer[_MT_co]: ...
     def get_serializer_class(self) -> type[BaseSerializer[_MT_co]]: ...
-    def get_serializer_context(self) -> dict[str, Any]: ...
+    def get_serializer_context(self) -> Mapping[str, Any]: ...
```
intgr, line 43: Maybe for follow-up: this return could be `TypedDict` already.

intgr (approved):
Thanks!

terencehonles:
This is actually constraining in a different way, the original implementation reflected that it was mutable and also had the same interface as a `dict`. This probably should have been a lower bound instead of changing to a protocol which mostly overlaps but also is stricter than the runtime implementation.

sevdog:
It should be changed into `MutableMapping` or any override of `get_serializer_context` method will get `Unsupported target for indexed assignment` error.

terencehonles:
@sevdog the linked PR #993 handles that issue

PR #964, opened 2026-04-13
Clean up and improve `rest_framework.fields`

emmanuel-ferdman:
### PR Summary
This PR cleans up `rest_framework.fields` for DRF 3.17 - adds `BigIntegerField`, removes the deprecated `NullBooleanField`, replaces all `Incomplete` types, fixes several wrong attribute and return types, adds missing instance attributes and method overrides, and makes defaults explicit.


scripts/stubtest/allowlist.txt
```diff
@@ -25,9 +25,16 @@ rest_framework.authtoken.migrations.0004_alter_tokenproxy_options
 # https://github.com/typeddjango/djangorestframework-stubs/issues/42
 rest_framework.fields.empty
 
+# Stub-only Protocol for iter_options() return type (no runtime class)
+rest_framework.fields.Option
+
+# Stub-only Protocol for ImageField typing (no runtime class)
+rest_framework.fields.SupportsToPython
```
intgr: I think we can get rid of these by using `@type_check_only` decorator and/or `_` prefix to the protocol class

rest_framework-stubs/serializers.pyi
intgr, line 40: I think `BigIntegerField` should also be re-exported here?

intgr (approved):
Thanks, a few questions. Everything else looks correct.

rest_framework-stubs/serializers.pyi
emmanuel-ferdman: Thanks, updated both :)

intgr (approved):
Awesome, thanks!

Issue #827, opened 2025-09-18
Incorrect type when creating `Serializer` with `many=True`


intgr:
TL;DR: `Serializer` behavior with `many=True` is funky. djangorestframework-stubs `3.16.3` changed the behavior of `ModelSerializer.instance` attribute. **If this caused a regression in your use case, please comment below.**

## Problem

```pycon
>>> from rest_framework.serializers import *
>>> class MySer(Serializer):
...     a = CharField(10)
...
>>> type(MySer())
<class '__main__.MySer'>
>>> type(MySer(many=True))
<class 'rest_framework.serializers.ListSerializer'>
```

Django REST Framework contains a particular trick: when creating an instance of any class that derives from `Serializer` with the parameter `many=True`, then the **constructor** does not actually return an instance of the class itself, but wraps the instance with `ListSerializer`. This is done using Python's magic `__new__()` method:

https://github.com/encode/django-rest-framework/blob/0f576223f22e162b3ef72247883e243a47aa36af/rest_framework/serializers.py#L123-L128

Currently, this cannot be accurately described for the `mypy` type checker, because the return types from `__new__()` are constrained: they can only be subclasses of the "self" type:

* https://github.com/python/mypy/issues/15182

## ModelSerializer `many=True` work-around was removed

Up until version `3.16.2`, djangorestframework-stubs contained a hack in the `ModelSerializer` class. The `instance` field was typed as `Model | Sequence[Model]`, to acommodate for the fact that `ModelSerializer` could be instanciated with either `many=True` or `many=False`.

This hack was removed in version `3.16.3` -- PR #719:

* There were multiple questions about this behavior and multiple requests to remove it.
* With the "union" approach, `instance` field type is compatible with **neither** `Model` nor `list[Model]`, requiring an override in both cases.
* I expect that in accessing `instance` of models with `many=True` is not a common use case -- it's a helpful convenience mostly for *nested* serializers.

**‼️ However, if you had a real use that was broken by version `3.16.3`, please share it in a comment below.**

PR #968, opened 2026-04-18
Recognize many=True serializer constructors as ListSerializer

federicobond:
# I have made things!

At runtime, `SomeSerializer(many=True, ...)` is swapped for a ListSerializer via `many_init`, but the stubs type it as the original serializer, so valid calls are rejected and downstream attributes lie about their types.

 This adds a `get_function_signature_hook` on every registered serializer class that, when many=True is literally passed:

  - accepts the ListSerializer-only kwargs min_length and max_length;
  - rewrites the constructor's return type to ListSerializer[Collection[_IN]], extracting _IN by mapping the child serializer's instance to BaseSerializer in its MRO.

### DIFF
--- a/mypy_drf_plugin/main.py
+++ b/mypy_drf_plugin/main.py
@@ -36,7 +36,7 @@ class NewSemanalDRFPlugin(Plugin):
     @override
     def get_function_signature_hook(self, fullname: str) -> Callable[[FunctionSigContext], CallableType] | None:
         if fullname in self._get_currently_defined_serializers():
-            return serializers.add_list_serializer_kwargs_when_many
+            return serializers.transform_serializer_constructor_when_many
         return None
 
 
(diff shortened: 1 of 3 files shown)


intgr:
Thanks! Sorry for the long delay in responding.

This is a known issue, also documented at https://github.com/typeddjango/djangorestframework-stubs/issues/827

Ideally, it would also be possible to express this with an overload of `__new__() -> ListSerializer` return type.

Problem is, right now mypy constrains the `__new__` return type to be a subclass of `Self`. But there's an open issue to relax this constraint on mypy, with lots of users requesting it. I suspect mypy upstream would accept to change it:

* https://github.com/python/mypy/issues/15182

I'd prefer to avoid maintaining a mypy plugin if possible. Would you be up to implementing this in mypy instead?

federicobond:
Will take a look and see what I can do!

intgr:
Great, thanks. If you see any reason why it wouldn't work then I can reconsider

intgr:
@federicobond Not sure if you're aware, the blocker was solved in mypy: https://github.com/python/mypy/pull/21441

federicobond:
Amazing. I was not aware!

PR #993, opened 2026-06-08
Update Serializer to capture context type

### LINKED ISSUE #986: get_serializer_context() return type should be dict, not Mapping
# Bug report
  
  ## What's wrong

  `GenericAPIView.get_serializer_context()` is typed as returning `Mapping[str, Any]`, which causes a type error when subclasses follow the common Django REST Framework pattern of calling `super()` and then adding keys to the returned  context:

  ```python
  def get_serializer_context(self):
      context = super().get_serializer_context()
      context['new_key'] = True  # Error here
      return context
  ```

  mypy reports:
  ```
  Unsupported target for indexed assignment ("Mapping[str, Any]")  [index]
  ```

  `Mapping` is read-only by design, so typed assignment to its keys is not allowed. Since `get_serializer_context()` implementations actually return a plain mutable `dict`, the return type should reflect that.

  ## How is that should be

  `get_serializer_context()` should be typed as returning `dict[str, Any]` instead of `Mapping[str, Any]`. This matches the actual runtime return type and allows callers to mutate the context without needing a `dict()` cast or `cast()`
  call:

  ```python
  def get_serializer_context(self):
      context = super().get_serializer_context()  # dict[str, Any]
      context['new_key'] = True  # No error
      return context
  ```

  Note: accepting a `Mapping` type as *input* to serializer `context=` is a separate and reasonable change (see #636), but the *return type* of `get_serializer_context()` should remain `dict`.

  ## System information

  - OS: Linux 7.0.5
  - `python` version: 3.14.4
  - `django` version: 6.0.5
  - `mypy` version: 1.20.0
  - `django-stubs` version: 6.0.4

terencehonles:
# I have made things!

This changes fixes the Serializer context no longer being mutable and instead binds the context to the value passed in on creation.

## Related issues

fixes a regression in https://github.com/typeddjango/djangorestframework-stubs/pull/946 (see: https://github.com/typeddjango/djangorestframework-stubs/pull/946#issuecomment-4648366288)

Closes https://github.com/typeddjango/djangorestframework-stubs/issues/986


rest_framework-stubs/serializers.pyi
```diff
@@ -61,24 +61,33 @@ from rest_framework.relations import PrimaryKeyRelatedField as PrimaryKeyRelated
...
 LIST_SERIALIZER_KWARGS_REMOVE: Sequence[str]
 ALL_FIELDS: str
 
+_VT = TypeVar("_VT", bound=APIView)  # View Type
+_CT = TypeVar("_CT", bound=_ContextType, default=_ContextType)  # Context Type
 _MT = TypeVar("_MT", bound=Model)  # Model Type
 _IN = TypeVar("_IN")  # Instance Type
 
-class BaseSerializer(Field[Any, Any, Any, _IN]):
+class _ContextType(TypedDict, Generic[_VT], extra_items=Any):
```
terencehonles: Mypy currently doesn't support extra items (see https://github.com/python/mypy/issues/18176) but this is the intent and the tests can just be updated when it does support it.

terencehonles: This doesn't seem to be working even if importing from typing_extensions, so I'm going to leave this off.

rest_framework-stubs/serializers.pyi
```diff
@@ -215,7 +224,7 @@ class ModelSerializer(Serializer[_MT]):
         *,
         partial: bool = ...,
         many: bool = ...,
-        context: Mapping[str, Any] = ...,
+        context: _CT = ...,
```
intgr, line 230: Doesn't this cause other problems though?

If you pass a custom context `ModelSerializer(..., context={"somevalue": 123})` then type-checkers will complain that the argument doesn't satisfy bound `_ContextType` because it's missing keys `request`, `format`, `view`?

The test appears to be using `cast()` to work around it.

terencehonles: The lower bound could be lower, I just figured usually you'd be using `get_serializer_context`.

PR #1016, opened 2026-07-23
Fix `request.data` type to include `list[Any]` for JSON bulk payloads

alanjds:
`request.data` can be a list when the client POSTs a top-level JSON array and JSONParser is used (e.g. bulk create/update endpoints). The stub previously typed it as `dict[str, Any]` only, forcing `# type: ignore[unreachable]` on legitimate
`isinstance(request.data, list)` checks.

The DRF "proof" that `request.data` can be a `list` is on the comment link below

## Related issues

- https://github.com/encode/django-rest-framework/issues/7633#issuecomment-725914908


intgr (approved):
Thanks!

intgr:
Released https://github.com/typeddjango/djangorestframework-stubs/releases/tag/3.17.1

PR #1057, opened 2026-09-17
Allow OrderingFilter, SearchFilter title/description to be translated strings

emorozov:
This fixes a trivial error in stubs for filters. If I have a filter with translated title or description:

```py
from rest_framework import filters

class CustomOrderingFilter(filters.OrderingFilter):
    ordering_param = 'o'
    ordering_title = _('Сортировка')
    ordering_description = _('Сортировка по нескольким параметрам.')
```

type checking fails:
```
filters.py:13: error: Incompatible types in assignment (expression has type "_StrPromise", base class "OrderingFilter" defined the type as "str")  [assignment]
filters.py:14: error: Incompatible types in assignment (expression has type "_StrPromise", base class "OrderingFilter" defined the type as "str")  [assignment]
```

titles and descriptions are actually lazily translations in DRF.

Haven't found this issue in the existing issues list.


sobolevn (approved):
Thanks!

PR #1058, opened 2026-09-18
Allow int keys in exception details and `ListSerializer.errors`

### LINKED ISSUE #1056: `_Detail` / `_APIExceptionInput` reject int-keyed mappings, but DRF raises `ValidationError` with index-keyed dicts
# Bug report

## What's wrong

`_Detail` and `_APIExceptionInput` in [`rest_framework-stubs/exceptions.pyi`](https://github.com/typeddjango/djangorestframework-stubs/blob/master/rest_framework-stubs/exceptions.pyi) only allow **string-keyed** mappings:

```python
_Detail: TypeAlias = ErrorDetail | list[_Detail] | dict[str, _Detail]
_APIExceptionInput: TypeAlias = (
    _Detail | StrOrPromise | Sequence[_APIExceptionInput] | Mapping[str, _APIExceptionInput] | None
)

class ValidationError(APIException):
    detail: list[_Detail] | dict[str, _Detail]
```

DRF itself raises `ValidationError` with **`int`-keyed** (list-index-keyed) mappings in at least two places, so the real type is `Mapping[int, _Detail]` / `Mapping[int, _APIExceptionInput]` as well.

### 1. `ListField.run_child_validation` — `rest_framework/fields.py`

https://github.com/encode/django-rest-framework/blob/3.18.0/rest_framework/fields.py#L1711-L1725

```python
def run_child_validation(self, data):
    result = []
    errors = {}

    for idx, item in enumerate(data):
        try:
            result.append(self.child.run_validation(item))
        except ValidationError as e:
            errors[idx] = e.detail          # <-- int key
        except DjangoValidationError as e:
            errors[idx] = get_error_detail(e)  # <-- int key

    if not errors:
        return result
    raise ValidationError(errors)
```

This has been the case for a long time (present at least since DRF 3.10).

### 2. `ListSerializer.to_internal_value` — `rest_framework/serializers.py`

https://github.com/encode/django-rest-framework/blob/3.18.0/rest_framework/serializers.py#L700-L712

```python
ret = []
errors = {}

for index, item in enumerate(data):
    try:
        validated = self.run_child_validation(item)
    except ValidationError as exc:
        errors[index] = exc.detail   # <-- int key
    else:
        ret.append(validated)

if errors:
    raise ValidationError(errors)
```

This one is **new in DRF 3.18.0** — it changed from `errors.append(exc.detail)` (a `list`) to an index-keyed `dict` in [encode/django-rest-framework#9837](https://github.com/encode/django-rest-framework/pull/9837). DRF 3.17.0 still used the list form. Since `djangorestframework-stubs` 3.18.x targets DRF 3.18.0, the stubs are now out of date with the version they support.

Note that `_get_error_details` passes dict keys through untouched (`{key: _get_error_details(value, default_code) for key, value in data.items()}`), so the `int` keys survive into `exc.detail`.

### Runtime evidence (DRF 3.18.0)

```python
>>> f = serializers.ListField(child=serializers.IntegerField())
>>> f.run_validation([1, 'x', 3, 'y'])
ValidationError: {1: [ErrorDetail(string='A valid integer is required.', code='invalid')],
                  3: [ErrorDetail(string='A valid integer is required.', code='invalid')]}

>>> class S(serializers.Serializer):
...     n = serializers.IntegerField()
>>> s = S(data=[{'n': 1}, {'n': 'bad'}], many=True)
>>> s.is_valid()
False
>>> s.errors
{1: {'n': [ErrorDetail(string='A valid integer is required.', code='invalid')]}}
>>> [type(k).__name__ for k in s.errors]
['int']
```

### Failing test cases

**Write side** — constructing the exception the way DRF does is rejected:

```python
from rest_framework.exceptions import ValidationError

errors: dict[int, list[str]] = {0: ["A valid integer is required."]}
raise ValidationError(errors)
```

```
error: Argument 1 to "ValidationError" has incompatible type "dict[int, list[str]]"; expected "_APIExceptionInput"  [arg-type]
```

**Read side** — handling the exception gives the wrong key type, silently:

```python
from rest_framework.exceptions import ValidationError

def handle(exc: ValidationError) -> None:
    if isinstance(exc.detail, dict):
        for index, detail in exc.detail.items():
            reveal_type(index)   # Revealed type is "str"  -- actually int at runtime
            print(index + 1, detail)
```

```
note: Revealed type is "str"
error: Unsupported operand types for + ("str" and "int")  [operator]
```

The read side is the worse of the two: the stubs don't just reject valid code, they actively assert a key type that is wrong at runtime, so code that correctly treats the key as an index is flagged as an error while code that treats it as a field name type-checks fine and then breaks.

### Related: `ListSerializer.errors`

The same DRF 3.18.0 change also makes [`rest_framework-stubs/serializers.pyi`](https://github.com/typeddjango/djangorestframework-stubs/blob/master/rest_framework-stubs/serializers.pyi) wrong for `ListSerializer.errors`, which is declared as:

```python
class ListSerializer(BaseSerializer[_IN]):
    @property
    @override
    def errors(self) -> ReturnList: ...
```

but at runtime returns `ReturnDict` or `ReturnList` depending on the error shape — https://github.com/encode/django-rest-framework/blob/3.18.0/rest_framework/serializers.py#L810-L819:

```python
@property
def errors(self):
    ret = super().errors
    if isinstance(ret, list) and len(ret) == 1 and getattr(ret[0], 'code', None) == 'null':
        detail = ErrorDetail('No data provided', code='null')
        ret = {api_settings.NON_FIELD_ERRORS_KEY: [detail]}
    if isinstance(ret, dict):
        return ReturnDict(ret, serializer=self)
    return ReturnList(ret, serializer=self)
```

With the reproducer above, `type(s.errors)` is `ReturnDict`, while `reveal_type(s.errors)` is `ReturnList[Any]`.

## How is that should be

The key types should admit `int` alongside `str`. Something along these lines:

```python
_DetailKey: TypeAlias = str | int
_Detail: TypeAlias = ErrorDetail | list[_Detail] | dict[_DetailKey, _Detail]
_APIExceptionInput: TypeAlias = (
    _Detail | StrOrPromise | Sequence[_APIExceptionInput] | Mapping[_DetailKey, _APIExceptionInput] | None
)
_ErrorCodes: TypeAlias = str | list[_ErrorCodes] | dict[_DetailKey, _ErrorCodes] | None
_ErrorFullDetails: TypeAlias = _FullDetailDict | list[_ErrorFullDetails] | dict[_DetailKey, _ErrorFullDetails]

class ValidationError(APIException):
    detail: list[_Detail] | dict[_DetailKey, _Detail]
```

`_ErrorCodes` and `_ErrorFullDetails` need the same treatment, since `_get_codes` and `_get_full_details` also rebuild the mapping with the original keys.

And in `serializers.pyi`:

```python
class ListSerializer(BaseSerializer[_IN]):
    @property
    @override
    def errors(self) -> ReturnDict | ReturnList: ...
```

One thing worth deciding as part of this: on DRF `master` the list-based format is still reachable via the new `LIST_SERIALIZER_ERRORS_AS_DICT` setting ([encode/django-rest-framework#10027](https://github.com/encode/django-rest-framework/pull/10027)), which defaults to `True` and is deprecated for removal in DRF 3.20. So for now both shapes are possible for `ListSerializer`, which argues for the union rather than swapping `str` for `int`. (`ListField` is unconditionally `int`-keyed and unaffected by that setting.)

I'm happy to open a PR if the proposed direction looks right.

## System information

- OS: macOS 15 (Darwin 24.6.0), arm64
- `python` version: 3.13.8
- `django` version: 5.2.14
- `mypy` version: 2.3.1
- `django-stubs` version: 5.2.9
- `djangorestframework` version: 3.18.0
- `djangorestframework-stubs` version: 3.16.9 (the definitions quoted above are from current `master`, i.e. 3.18.1, and are unchanged)

JPDSousa:
Fixes #1056.

## What

DRF keys error mappings by list index, not only by field name, but `_Detail` / `_APIExceptionInput` only allowed `str` keys.

- `ListField.run_child_validation` has raised `ValidationError({index: detail})` since at least DRF 3.10.
- `ListSerializer.to_internal_value` does the same as of DRF 3.18.0 (encode/django-rest-framework#9837 changed it from an `errors.append(...)` list).

`_get_error_details` passes mapping keys through untouched, so those `int` keys survive into `.detail`, `.get_codes()` and `.get_full_details()`. Verified against the pinned `djangorestframework==3.18.0`:

```
ListField: {1: [ErrorDetail(string='A valid integer is required.', code='invalid')], 3: [...]} ['int', 'int']
errors: {1: {'n': [ErrorDetail(string='A valid integer is required.', code='invalid')]}} ReturnDict ['int']
```

The read side was the worse half: the stubs didn't just reject valid code, they asserted `str` for a key that is `int` at runtime, so code that correctly treated the key as an index was flagged while code that treated it as a field name type-checked and then broke.

## Changes

`rest_framework-stubs/exceptions.pyi`:

- New `_DetailKey: TypeAlias = str | int`, used for the mapping keys in `_Detail`, `_ErrorCodes`, `_ErrorFullDetails` and `ValidationError.detail`.
- `_APIExceptionInput` gains a `Mapping[int, _APIExceptionInput]` member.

  Note this is deliberately **not** a single `Mapping[_DetailKey, _APIExceptionInput]`: `Mapping` is invariant in its key type, so that form rejects both `dict[str, ...]` and `dict[int, ...]` arguments and would have been a regression for existing `dict[str, ...]` callers. Spelling the two key types out as separate union members keeps both working. There's a comment in the stub recording this.

`rest_framework-stubs/serializers.pyi`:

- `ListSerializer.errors` is now `ReturnDict | ReturnList`. The same DRF 3.18 change means it returns a `ReturnDict` for index-keyed errors (and for the `'No data provided'` non-field case), while previously the stub claimed `ReturnList` unconditionally. The list shape stays reachable on DRF `master` via the new `LIST_SERIALIZER_ERRORS_AS_DICT` setting (encode/django-rest-framework#10027), which argues for the union rather than swapping one for the other.

## Tests

- `tests/typecheck/test_exceptions.yml`: new `test_exception_input_index_keyed` case covering index-keyed dict literals, a pre-typed `dict[int, list[str]]`, a pre-typed `dict[str, list[str]]` (the no-regression case) and a still-rejected bad value type.
- `tests/assert_type/exceptions.py`: asserts the iterated key type of `ValidationError.detail` is `str | int`, plus the two write-side calls.
- `tests/assert_type/serializers.py`: asserts `ListSerializer.errors` is `ReturnDict[Any, Any] | ReturnList[Any]`.
- Two expected messages in the existing `test_exception_input` case changed. With two `Mapping` members in the union, mypy reports a whole-argument mismatch instead of the narrower `Dict entry 0` / `List item 1` diagnostics. The calls are still errors, only the message is less specific.

`pytest`, `mypy rest_framework-stubs`, `mypy tests` and `scripts/stubtest.sh` all pass locally.

## Not included

`as_serializer_error` is declared as returning `dict[str, list[ErrorDetail]]` and has the same key-preserving behaviour, so its key type is wrong in the same way — and its value type is too, since a nested `Mapping` value is passed through as-is. That felt like a separate change; happy to fold it in here or open a follow-up, whichever you prefer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rest_framework-stubs/serializers.pyi
```diff
@@ -186,9 +186,11 @@ class ListSerializer(BaseSerializer[_IN]):
     @property
     @override
     def data(self) -> ReturnList: ...
+    # Since DRF 3.18 `to_internal_value` collects child errors in a list-index-keyed dict, so `.errors`
+    # is a `ReturnDict` for that shape (and for non-field errors) and a `ReturnList` otherwise.
     @property
     @override
-    def errors(self) -> ReturnList: ...
+    def errors(self) -> ReturnDict | ReturnList: ...
```
intgr: What's "otherwise" - when is it a `ReturnList`?

tests/assert_type/serializers.py
```diff
@@ -163,3 +163,10 @@ class SomeContext(TypedDict):
 
 context_model_ser: serializers.ModelSerializer = serializers.ModelSerializer(None, context=context)
 assert_type(context_model_ser.context, Mapping[str, Any])
+
+
+# case: test_list_serializer_errors_may_be_index_keyed
+# Since DRF 3.18 `ListSerializer.to_internal_value` collects child errors in an index-keyed dict,
+# so `.errors` is a `ReturnDict` for that shape and a `ReturnList` otherwise.
+list_ser: serializers.ListSerializer = serializers.ListSerializer(child=serializers.Serializer())
+assert_type(list_ser.errors, ReturnDict[Any, Any] | ReturnList[Any])
```
intgr: That's obvious from stubs and doesn't need a test.

rest_framework-stubs/exceptions.pyi
```diff
@@ -11,20 +11,31 @@ class ErrorDetail(str):
     code: str | None
     def __new__(cls, string: str, code: str | None = ...) -> Self: ...
 
-_Detail: TypeAlias = ErrorDetail | list[_Detail] | dict[str, _Detail]
+# DRF keys error mappings by list index, not just by field name: `ListField.run_child_validation`
+# and (since DRF 3.18) `ListSerializer.to_internal_value` both raise `ValidationError({index: detail})`.
+# `_get_error_details` passes mapping keys through untouched, so those `int` keys reach `.detail`.
+_DetailKey: TypeAlias = str | int
+_Detail: TypeAlias = ErrorDetail | list[_Detail] | dict[_DetailKey, _Detail]
```
intgr: Note there are two different constructs:

* `dict[str | int, T]` (ints and strings may be intermixed)
* `dict[str, T] | dict[int, T]` (uniformly typed)

Is there good reason to use former here, and latter in `_APIExceptionInput`?

tests/typecheck/test_exceptions.yml
```diff
@@ -18,5 +18,22 @@
...
+    ValidationError({0: ['A valid integer is required.']})
+    ValidationError({0: {'field': ErrorDetail('required', code='required')}})
+
+    index_errors: dict[int, list[str]] = {0: ['A valid integer is required.']}
+    ValidationError(index_errors)
+
+    field_errors: dict[str, list[str]] = {'field': ['A valid integer is required.']}
+    ValidationError(field_errors)
+
+    ValidationError({0: 1})  # E: Argument 1 to "ValidationError" has incompatible type "dict[int, int]"; expected "_APIExceptionInput"  [arg-type]
```
intgr: We've been migrating to assert_type tests instead of typecheck for simpler tests. The `E:` case can't be covered there, but doesn't provide enough value IMO.

Actually would be nice to have assert_type tests that cover all possible valid inputs to `_APIExceptionInput`

rest_framework-stubs/exceptions.pyi
JPDSousa: You're right 👍 there's never `int | str` in the dict keys

rest_framework-stubs/serializers.pyi
JPDSousa: Clarified in the docstring:
* `LIST_SERIALIZER_ERRORS_AS_DICT = False` still serializes errors as a list
* There's this weird corner case where, in a successful validation of a ListSerializer, `serializer.errors == []`

intgr (approved):
Looks good, thanks!

PR #1059, opened 2026-09-19
Fix incomplete types in templatetags

emmanuel-ferdman:
### PR Summary
This PR replaces all `Incomplete` return types in `rest_framework` - template tags and filters now return `str`, `highlight_code` returns `CodeNode`, `items` returns `ItemsView[Any, Any] | list[Any]` (matches `dict.items()`), and `smart_urlquote_wrapper` returns `str | None` (since it catches `ValueError`).


intgr (approved):
Thanks, LGTM

